PR #2, opened 2016-08-25
Add use strict for Node < v6.x users

radum:
Block-scoped declarations (let, const, function, class) not yet supported outside strict mode for Node < v6.x.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,5 +28,8 @@
   "bugs": {
     "url": "https://github.com/paulirish/pwmetrics/issues"
   },
-  "homepage": "https://github.com/paulirish/pwmetrics#readme"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,5 +28,8 @@
   "bugs": {
     "url": "https://github.com/paulirish/pwmetrics/issues"
   },
-  "homepage": "https://github.com/paulirish/pwmetrics#readme"
+  "homepage": "https://github.com/paulirish/pwmetrics#readme",
+  "engines" : {
+    "node" : ">= 4.4.6"
```
hemanth: I was here to add that, cool!

paulirish:
thx!

PR #11, opened 2016-11-19
Add expectation options

### LINKED ISSUE #10: Add possibility to use it with CI
I would really like to have some option like `maxTimeSinceNavSrart`, value of which means speed expectation.
Using this option `pwmetrics` might throw some error `Time (ms) since navigation start more then you expected, so improve your page perf`.
This approach could be used for CI.

denar90:
Resolves #10

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -20,7 +20,15 @@ if (!url || flags.help) {
   if (!flags.help) console.error('No url entered.');
   console.error('Usage:');
   console.error('    pwmetrics http://goat.com');
-  console.error('    pwmetrics --json http://goat.com');
+  console.error('    pwmetrics http://goat.com --json  Reports json details to stdout.');
+  console.error('    pwmetrics http://goat.com --runs=n  Does n runs (eg. 3, 5), and reports the median run\'s numbers.');
+  console.error('    pwmetrics http://goat.com --ttfcp  Expected First Contentful Paint');
+  console.error('    pwmetrics http://goat.com --ttfmp  Expected Meaningful Contentful Paint');
+  console.error('    pwmetrics http://goat.com --psi  Expected Perceptual Speed Index');
+  console.error('    pwmetrics http://goat.com --fv  Expected First Visual Change');
+  console.error('    pwmetrics http://goat.com --vc  Expected Visually Complete 100%');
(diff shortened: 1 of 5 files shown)


paulirish:
cc @brendankenny as this is similar to smokehouse
cc @samccone

paulirish:
@denar90 thx for much for submitting this. finally have some time to take a look and write down where my head is at..

first, take a look at https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/test/smokehouse/smokehouse.js and try it in lighthouse with `npm run smoke`

You can see their expectations set in https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/test/smokehouse/pwa-expectations.js .. those are only handling the booleans, whereas you want to handle numbers though.

it does something similar but it's implemented separately from the core, which is something I think we should follow here. I'd like most of this to be implemented in a separate file.

so here's one possible way it could look for us:

### expectations.js

```js
module.exports = {
  url: 'http://localhost/awesomesite/',
  metrics: {
    ttfmp: {
      warn: '>=3000',
      error: '>=5000'
    },
    psi: {
      warn: '>=1500',
      error: '>=3200'
    }
  }
}
```

This'll trigger a run with the provided URL and then compare the results with these values.  Something gets logged on both warn & error and if there's and `error` then it exits with `1`.

Maybe that's triggered with `pwmetrics --expectations=expectations.js` ? 

just spitballing here, but would that work for you?

denar90:
Thanks for feedback. Definitely better approach then mine, I like it. 
As thought, this config can be part of `package.json` in future.

P.S. Need to find some spare time for implementing it.

bin/cli.js
```diff
@@ -16,11 +17,22 @@ argv
     flags[flagKey] = keyValue[1] || true;
   });
 
+let url;
```
denar90: I think in future we can replace all checks here by some validator. It will make cli file more clear.

paulirish:
Oh this is nice. Will take a look later today.

paulirish:
Nice stuff! 

I gave you a PR on your fork with some proposals: https://github.com/denar90/pwmetrics/pull/1

denar90:
I can add some tests for critical functionality in another PR and then we can publish a new version, ah?

paulirish:
Sounds good!

PR #14, opened 2016-12-13
pass through flags to lighthouse from pwmetrics CLI

paulirish:
For example:

```sh
   pwmetrics --runs=3 --disableDeviceEmulation "chrome://history/" --save-assets
```

wdyt

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ class PWMetrics {
   }
 
   recordLighthouseTrace() {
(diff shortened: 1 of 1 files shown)


denar90:
@samccone do you wanna merge it?

paulirish:
Yup, just wanted your +1 :)

It's a little weird since its the camelCase opts but i'll live.

PR #17, opened 2017-01-04
Minor improvements

denar90:
- fixed messages with prefixes
 - `var` -> `const`
 - consistent `url` example for docs


paulirish:
looks good! thanks

PR #18, opened 2017-01-05
experimental: post results to google sheets

paulirish:
#### Results:
image
image
Live link: https://docs.google.com/spreadsheets/d/1-84NpWXXBNmwkmx4tmWNzR7xnVYRhIBgVxGePA4jsZk/edit?usp=sharing

#### Instructions: 

(There's some overly manual stuff right now...)

1. setup google developer project and get credentials. ([everything in step 1 here](https://developers.google.com/sheets/api/quickstart/nodejs#step_1_turn_on_the_api_name))
2. take the client_secret and place in `lib/gsheets/`
3. run `node lib/gsheets/gsheets.js` and follow instructions for authenicating
4. make a copy of [this spreadsheet](https://docs.google.com/spreadsheets/d/1-84NpWXXBNmwkmx4tmWNzR7xnVYRhIBgVxGePA4jsZk/edit?usp=sharing), remove rows 2-N.
5. copy the ID of the spreadsheet into the `spreadsheetId` in `gsheets.js`

then run `pwmetrics http://paulirish.com --submit`

with luck, you'll be in business.
What's really cool is seeing the chart update the second that new results are posted. :D

PR #20, opened 2017-01-07
Sheets improves

denar90:
I added some options which can be passed from config file, so now we can ignore 5-th step from instruction list https://github.com/paulirish/pwmetrics/pull/18#issue-198892447

Next think is how to get rid of 
```
3. run node lib/gsheets/gsheets.js and follow instructions for authenicating
4. make a copy of this spreadsheet, remove rows 2-N.
```
Do you think this one https://github.com/google/google-api-nodejs-client/ can be helpful for case 3?

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -3,8 +3,7 @@
 // Licensed under the Apache License, Version 2.0. See LICENSE
 
 const PWMetrics = require('../lib/index');
-const { getExpectationsConfig } = require('../lib/expectations');
-const { getSheetsConfig } = require('../lib/gsheets/options-adapter');
+const { getConfigFromFile } = require('../lib/utils/fs');
 const { getMessageWithPrefix } = require('../lib/utils/messages');
 
 const argv = process.argv.slice(2);
@@ -18,18 +17,9 @@ argv
     flags[flagKey] = keyValue[1] || true;
   });
 
-// @todo refactor
-let url;
-if (flags.expectations) {
-  const expectationsConfig = getExpectationsConfig(flags.expectations);
-  url = expectationsConfig.url;
-  flags.metrics = expectationsConfig.metrics;
-} else if (flags.submit) {
-  flags.sheet = getSheetsConfig(flags.submit);
-  delete flags.submit;
-}
+const config = Object.assign({}, getConfigFromFile(flags.config), { flags: flags });
 
-url = url || argv.filter(f => !f.startsWith('-')).shift();
+const url = config.url || argv.filter(f => !f.startsWith('-')).shift();
 
 if (!url || flags.help) {
   if (!flags.help) console.error(getMessageWithPrefix('ERROR', 'NO_URL'));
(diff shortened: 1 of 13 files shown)


bin/cli.js
```diff
@@ -17,15 +18,19 @@ argv
...
 if (flags.expectations) {
-  const expectationsConfig = getConfig(flags.expectations);
+  const expectationsConfig = getExpectationsConfig(flags.expectations);
   url = expectationsConfig.url;
   flags.metrics = expectationsConfig.metrics;
-} else {
-  url = argv.filter(f => !f.startsWith('-')).shift();
+} else if (flags.submit) {
+  flags.sheet = getSheetsConfig(flags.submit);
+  delete flags.submit;
```
denar90: I'm not sure about this one. Maybe it should be `submit` object which contains 
```
sheets: {
  type: ...,
  options: {}
}
```

denar90: Another think. I'll just leave it here. It's about flags `submit`, `expectations`. Right now for both of them users have to pass file path as value. What if we can do it better, like
`pwmetrics --submit --config=./path-to-config`
`pwmetrics --expecations --config=./path-to-config`
if path is not passed then config will be read from `package.json`.
Format of config file might be
```js
{
  url: 'http://example.com/',
  expectations: {
    metrics: {
      ttfcp: {
        warn: '>=30',
        error: '>=5000'
      },
      ttfmp: {
        warn: '>=3000',
        error: '>=50'
      },
      psi: {
        warn: '>=1500',
        error: '>=32'
      }
    }
  },
  sheets: {
    type: 'GOOGLE_SHEETS',
    options: {
      spreadsheetId: '1Uo4a4KQTaC1FBDPMIJ6xLMmOd2yLyps-6jBC5x4x6I8',
      tableName: 'data'
    }
  }
}

```

lib/gsheets/gsheets.js
```diff
@@ -23,7 +21,7 @@ var TOKEN_PATH = TOKEN_DIR + 'sheets.googleapis.com-nodejs-pwmetrics.json';
 function authenticate() {
   return new Promise((resolve, reject) => {
     // Load client secrets from a local file.
-    fs.readFile('client_secret.json', 'utf8', function processClientSecrets(err, content) {
+    fs.readFile('./lib/gsheets/client_secret.json', 'utf8', function processClientSecrets(err, content) {
```
denar90: Nice idea. Do you suggest to have this property value as path? Like
```
{
 ...
  clientSecretDataPath: '/my-global-path/client_secret.json'
 ...
}

paulirish: eh. i'm thinking of like a `client_secret_contents` property and then just paste in the content of that original file.. so you just have one file total.

denar90: Make sense. Thanks 👍

paulirish:
merging this into the branch..

PR #23, opened 2017-01-10
Add missed barLabel

### LINKED ISSUE #22: No actual value of each bar in the graph is painted
Hi I ran into an possible issue. I am running `node 7.4.0` installed with homebrew on osx. I've installed `lighthouse@1.4.1` and `pwmetrics@1.2.3` globally with `npm install -g`. When i run either `pwmetrics` or with the `—runs=3` option I get an output like this. 

bildschirmfoto 2017-01-10 um 11 43 42

I get an rough idea about the dimension by the 10.000 value on the x-axis but the specific values are missing like in the example screenshot in the `readme.md` file. If you need any further informations let me know. Best regards r.

denar90:
`barLabel` was missed during chart moving in repo. I added changes due to https://github.com/samccone/cli-chart/pull/2/files

This should fix #22 
barlabel

### DIFF
--- a/lib/chart/bar.js
+++ b/lib/chart/bar.js
@@ -7,6 +7,7 @@ var Bar = module.exports = function(chart, opts) {
   this.size = opts.size;
   this.color = opts.color || defaults.color;
   this.label = opts.label;
+  this.barLabel = opts.barLabel;
 };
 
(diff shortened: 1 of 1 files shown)


paulirish:
thanks. not sure how i missed this.

PR #29, opened 2017-01-12
travis fix

denar90:


paulirish:
Thanks. I was looking at all the other travis yaml's on github and never saw this, so i figured it wasn't neccessary.. but i'm also not sure why the current is failing. :) 

thx

PR #30, opened 2017-01-15
Sheets finalizations

denar90:
I think after merging this we can try release `sheets` stuff. Looks like it will be major one 😃  We should definitely concentrate on tests after that.


lib/gsheets/gsheets.js
```diff
@@ -1,21 +1,25 @@
-var fs = require('fs');
-var readline = require('readline');
-var google = require('googleapis');
-var GoogleAuth = require('google-auth-library');
-var process = require('process');
+// Copyright 2016 Google Inc. All Rights Reserved.
```
denar90: @paulirish what do you think if we implement logic this file as a `Class` ? It will be easier to test + more functionality will be covered...

paulirish: as a class works for me. this was originally some copypaste code from their examples.. but yeah i'm a fan of cleaning it up. :)

lib/gsheets/gsheets.js
```diff
@@ -131,8 +140,9 @@ const formatValues = values => values.slice(3).join('\t');
  * @return {Promise}
  */
 function appendResults(auth, valuesToAppend, options) {
-  var sheets = google.sheets('v4');
-  const values = [valuesToAppend];
+  const sheets = google.sheets('v4');
+  // clone values to append
+  const values = valuesToAppend.slice();
```
paulirish: I'd rather do a JSONparse(jsonstringify(valuestoappend)) or Object.assign([], valuesToAppend) to clone instead.

lib/index.js
```diff
@@ -77,11 +76,10 @@ class PWMetrics {
     return lighthouse(this.url, lhOpts, perfConfig)
       .then(res => metrics.prepareData(res))
       .then(data => {
-        if (this.flags.submit) {
-          return this.appendResults(data);
```
paulirish: I had put it here so I would get results streaming into sheets continuously if i did --runs=100. With this move, it looks like it'd only report them all when the 100 finish, am i reading this right?

lib/gsheets/gsheets.js
paulirish: shall we do a class after the big PR lands?  or before. I'm OK either way.

paulirish:
@denar90 i'm merging all this into master to make things a bit more sane. can you do these changes in a followup?

paulirish:
squashed into a2b7ef934b1008d4181431e5ed2dcfc836c6d547 and cherrypicked that to master.

lib/index.js
denar90: @paulirish yeap, the reason I did it, to not spam sheets with each request (in your case it will be 100 requests). If you think it's ok to send data on each run, then we can bring it back.

Issue #32, opened 2017-01-17
Feature request: ability to display chart AND metrics errors/warnings


MickeyKay:
First off, fantastic project - I love it! Thanks a ton for working on this, and in general for all the great FE work you do.

Secondly, I have a feature request. At present, pwmetrics allows you to either 1. display a performance chart or 2. include expectation metrics and display errors/warnings, however you can't do both: https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L75. According to the logic here, a chart will only ever be output if no expectations are set.

My request is to add the ability to display the chart and any expectation errors/warnings.

denar90:
#33 merged

PR #33, opened 2017-01-17
Add default display of chart alongside expectations

MickeyKay:
Addresses https://github.com/paulirish/pwmetrics/issues/32

This PR enables default display of the chart when `expectations` are present, and adds a new `disableChart` option that can be used to output expectation errors and warnings alone.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -86,10 +86,11 @@ class PWMetrics {
   displayOutput(data) {
     if (this.flags.json) {
       return data;
-    } else if (this.flags.expectations) {
-      expectations.checkExpectations(data.timings, this.expectations.metrics)
     } else {
       this.showChart(data);
+      if (this.flags.expectations) {
+        expectations.checkExpectations(data.timings, this.expectations.metrics);
+      }
     }
     return data;
   }


lib/index.js
```diff
@@ -20,7 +20,6 @@ class PWMetrics {
   constructor(url, opts) {
     this.url = url;
     this.opts = opts;
-    this.opts.disableCpuThrottling = this.opts.disableCpuThrottling || false;
```
denar90: I don't think that we should remove this one.

lib/index.js
```diff
@@ -76,6 +75,9 @@ class PWMetrics {
     if (this.opts.json) {
       return data;
     } else if (this.opts.expectations) {
+    	if (!this.opts.disableChart) {
```
denar90: What do you think about using `showChart` option and make it true by default?

MickeyKay: For sure - love it!

lib/index.js
MickeyKay: Restored. Accidentally removed this in my haste.

lib/index.js
```diff
@@ -76,6 +77,9 @@ class PWMetrics {
     if (this.opts.json) {
       return data;
     } else if (this.opts.expectations) {
+      if ( this.opts.showChart && 'false' != this.opts.showChart ) {
```
MickeyKay: Not sure if this is the best way to do this, however it seems that setting `--showChart=false` in CLI passes the `showChart` option through as a string instead of a bool. I thought about automatically parsing this in `cli.js`, however I think this is a different task altogether, and it seems like there are plenty of packages out there to handle this if you decide to go that route. Would love to hear if you have a better suggestion on how to handle this more gracefully.

lib/index.js
```diff
@@ -21,6 +21,7 @@ class PWMetrics {
     this.url = url;
     this.opts = opts;
     this.opts.disableCpuThrottling = this.opts.disableCpuThrottling || false;
+    this.opts.showChart = (typeof this.opts.showChart !== 'undefined') ? (this.opts.showChart == 'true') : true;
```
MickeyKay: Not sure if this is the best way to do this, however it seems that setting `--showChart=false` in CLI passes the `showChart` option through as a string instead of a bool. I thought about automatically parsing this in `cli.js`, however I think this warrants a different PR altogether if you'd like to go in that direction. This tidbit forces a boolean value no matter what. Would love to hear if you have a better suggestion on how to handle this more gracefully.

MickeyKay:
@denar90 Updated per your suggestions. I posed one question above and would love to hear your thoughts.

lib/index.js
denar90: I see 2 possible solutions here
1. easy one like you did, just simplify a bit: `this.opts.showChart = (this.opts.showChart === 'false') ? false : true;`
2. manage it on `cli.js` level - https://github.com/paulirish/pwmetrics/blob/master/bin/cli.js#L17
```
flags[flagKey] = keyValue[1] || true;
flags[flagKey] = (flags[flagKey] === 'false') ? false : flags[flagKey];
```
I'd prefer second one.

lib/index.js
```diff
@@ -76,6 +77,9 @@ class PWMetrics {
     if (this.opts.json) {
       return data;
     } else if (this.opts.expectations) {
+      if (this.opts.showChart) {
+        this.showChart(data);
+      }
       expectations.checkExpectations(data.timings, this.opts.metrics)
```
denar90: Just one snit pick. @MickeyKay what do you think if we draw chart after expectations will be checked. Like
```
expectations.checkExpectations(data.timings, this.opts.metrics);
if (this.opts.showChart)  this.showChart(data);
```

paulirish:
what if we just always show the chart, even if we're running against expectations?

MickeyKay:
@paulirish this was my initial preference, hence the original (optional) `disableChart` option. That seems cleaner to me.

lib/index.js
MickeyKay: I actually prefer the existing arrangement, and here's why. Whenever I'm running any script, the final thing I always want to see in the output are errors/warnings. Reason being is that it's easier to lose track of said errors/warnings if they're earlier on in such output, and much less likely lost if they are the last piece of output.

Also, it seems to me like the logical order of things is 1. see timings 2. see errors/warnings resulting from these timings.

With that said, my preference would be to keep as is. Thoughts?

denar90:
@MickeyKay @paulirish SGTM.

MickeyKay:
@denar90 @paulirish Okay, after all is said and done this PR is now super simple and simply outputs the chart regardless.

denar90:
@MickeyKay thanks!

MickeyKay:
Oh wow. Merged already! Thanks 👍

PR #38, opened 2017-01-24
Sheets imps

denar90:
- `appendResults` and validation stuff moved into separate Class, some tests

@paulirish what do you think about using `TypeScript` in future? It hurts a bit to check all arguments for each method.


paulirish:
> @paulirish what do you think about using TypeScript in future? It hurts a bit to check all arguments for each method.

yup i'd be happy to move to TS

paulirish:
this PR looks great. sorry for letting it sit.

denar90:
No problem at all :)
Do you wanna merge it? Looks like smth wrong with gulp-example which is run with Travis...

denar90:
Fixed via https://github.com/paulirish/pwmetrics/commit/b3d2fc91b45c62266bb373a32046506789f2c57e

PR #40, opened 2017-01-25
Recipes

denar90:
- added recipes for `grunt` and `gulp`
 - ~running them on `travis`, kinda functional testing how it works. In future we can add more examples and run them.~ Locally it works fine but It breaks build - https://travis-ci.org/paulirish/pwmetrics/jobs/195163109 Need more time for this.


paulirish:
I think we should nuke the grunt recipe.. it's a bit dated nowadays. :/

Also, it looks like we have a bug that's breaking the travis usage. I suppose that is separate, but we should solve it.  you can run `node --trace-warnings` to have it log more detail about these promise rejection errors.

denar90:
Thanks, will try to trace the problem. 

About `grunt`, it's really not a mainstream right now, but the recipe for it, just for users who still use it in real projects and don't wanna break which works for them.  
If we take a look at [npm](https://www.npmjs.com/package/grunt), it's `1,766,457 downloads in the last month` right now. So it's kinda big number.
So I thought it can be useful. 
But I'm also fine with dropping it.

denar90:
I found where problem was. It happened because of missed chrome on Travis. So right now it works fine.

---
But we have 2 issues here:
1. Chart is not show at all - https://travis-ci.org/paulirish/pwmetrics/jobs/196148490#L682
2. Sometimes CI can fails because of `CriConnection:error` - https://travis-ci.org/paulirish/pwmetrics/jobs/196148490#L682.
I found solution in `smokehouse` - https://github.com/GoogleChrome/lighthouse/commit/227487a3d0e3300ff97dce08a1da026ce5f844a9. I'll try to implement smth similar.

What we have to decide in to close  this PR?
1. Remove or keep grunt (both 👍 )?
2. Do we wanna add [webpack](https://webpack.github.io/) and [brunch](http://brunch.io/) examples in scope of this PR or in future (I'm 👍  for future)?

paulirish:
lets nuke grunt.

I'm +1 on adding webpack example in the future.. no need to do it here.

denar90:
done 💣 💥

paulirish:
nice stuff!

PR #42, opened 2017-01-29
Handle CRI_TIMEOUT error

denar90:
Resolves problem with lighthouse `CriConnection:error` on CI.


paulirish:
The runLighthouse function is pretty tangled. I'm having a hard time following the logic just reading it through, so maybe we can try another approach?

Perhaps break out an innerRunLighthouse to keep the retry logic separate from the actual calls to the lighthouse module?

denar90:
Yeap it looks too complicated, I'll sit and think about simplification.

paulirish:
@brendankenny did note that we've never needed this timeout stuff for anything other than travis.  So if that's helpful we could try to isolate this and only do it in the travis case...

denar90:
Yeap, lets go this way.

paulirish:
yah. much happier with this. thanks!

Issue #44, opened 2017-01-30
Median Run output inconsistent with normal output


pedro93:
This is a minor issue but might create some confusion for some users.
When running pwametrics with the ```runs``` flag the final median output's labels and data are not in the same order as the rest:

![screen shot 2017-01-30 at 15 00 28](https://cloud.githubusercontent.com/assets/3718654/22427891/da911a08-e6fc-11e6-9738-89f7f60c5097.png)

Notice how the last graph's labels are:
 - Time to Interactive
 - Perceptual Speed Index
 - First Meaningful Paint
 - First Contentful Paint

And the run graphs (which match the standard ```$ pwmetrics {url}``` output):
 - First Contentful Paint
 - First Meaningful Paint
 - Perceptual Speed Index
 - Time to Interactive

Merely changing this [line](https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L105) to:
```js
let timings = data.timings;
```

(Basically remove the ```.reverse()```)
Appears to fix the issue.

denar90:
@pedro93 thanks, nice catch 👍  Do you wanna make a PR?

pedro93:
Its a very minor change, I can if you want :)

denar90:
Merged #46

PR #46, opened 2017-01-30
Median output

pedro93:
Addresses https://github.com/paulirish/pwmetrics/issues/44, inconsistent graph output between median and normal runs.

@denar90 This is my first PR, is it ok?


denar90:
Thanks 👍

Issue #48, opened 2017-01-31
Config file not working


pedro93:
I am trying to create a configuration file to use with pwmetrics however it is not being used by the library. I have a file performance-test.config.js that looks like this:
```js
module.exports = {
    runs: 5,
    flags: {
        expectations: true,
    },
    sheets: {
        // sheets configuration
    },
    expectations: {
        ttfmp: {
            warn: '>=3000',
            error: '>=5000',
        },
        tti: {
            warn: '>=5000',
            error: '>=15000',
        },
        ttfcp: {
            warn: '>=1500',
            error: '>=3000',
        },
        psi: {
            warn: '>=3000',
            error: '>=6000',
        },
    },
};
```

I'm executing pwmetrics like this:
```js
function runPerf(url, cb){
    const pwmetrics = path.join(__dirname, '../node_modules/pwmetrics/bin/cli.js');
    const perfConfig = require('../config/performance-test.config.js');
    const child = cp.exec('node ' + pwmetrics + ' ' + url + ' --config=' + perfConfig);
    child.stderr.pipe(process.stderr);
    child.stdout.pipe(process.stdout);
    child.on('close', (code) => {
        cb(code);
    });
}
```

I think the issue is with how the [cli](https://github.com/paulirish/pwmetrics/blob/master/bin/cli.js) reads the options from the command line. Is there an option to override every command line option with the data from a json?

denar90:
Do you use latest version (1.2.4) or master? API you wanna use is in master branch and it hasn't been release yet.

---
Tips:
 - use `npm i paulirish/pwmetrics --save-dev` (until we release new awesome stuff)
 - use API for modules described in doc - https://github.com/paulirish/pwmetrics#api instead of executing cli.

E.g.
`my_module.js`
```js
const PWmetrics = require('pwmetrics');
const perfConfig = require('path-to-my-config/config.js');
....
function runPerf(url, cb){
  const pwmetrics = new PWmetrics(url, pwmetricsConfig);
  pwmetrics.start().then(cd);
}
```

`perfConfig.js`
```js
module.exports = {
    //flags - the place where cli options can be put
    flags: {
        expectations: true,
        runs: 5
    },
    sheets: {
        // sheets configuration
    },
    expectations: {
        ttfmp: {
            warn: '>=3000',
            error: '>=5000',
        },
        tti: {
            warn: '>=5000',
            error: '>=15000',
        },
        ttfcp: {
            warn: '>=1500',
[...]

Issue #47, opened 2017-01-30
Reduce output when doing runs


pedro93:
Consider only outputting the median graph data after all runs conclude.
Reason: Less information outputted to console and less cognitive load for the user. 

Personally if I ask pwmetrics to execute a number of times I'm more interested in seeing the median values than any 1 execution. 

Instead of displaying the chart of every run, pwmetrics could generate a file (JSON, CSV, whatever) with the data of every run but output to the console the chart for the median values. I understand there is a "sheets" configuration option but that requires configuring a google sheets and time to setup up. I propose a more "local" approach.

I could make a PR for this if the maintainers of this repository find it worthwhile to have this feature. To simply reduce the terminal output, the changes would occur here:
https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L84
By simply changing the if statement to:
```js  
if (!this.flags.submit && this.runs <= 1) {
    this.displayOutput(data);
}
```

Thoughts? @denar90 @paulirish

paulirish:
Yup i support doing something different here.

Minimal visual output seems good to me.

pedro93:
Do you have any suggestions? I can make a PR that would save a json or csv with the intermediate runs.

denar90:
We have related issue here https://github.com/paulirish/pwmetrics/issues/34
I'm ok for now just reducing number of graphics which will be shown.

denar90:
Looks like solved.

Issue #53, opened 2017-02-07
Remove duplicated pwmetrics object properties and merging yargs support


pedro93:
Whilst reading through the project I noticed the [pwmetrics object](https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L22) contains a property ```opts``` which is not used anywhere else, and will in fact duplicate the object size in memory. 

If you agree @paulirish I would like to do a code cleanup for this and some other minor fixes across files including adding the [yargs PR](https://github.com/paulirish/pwmetrics/pull/51), improve upon it and potencially add support for https://github.com/paulirish/pwmetrics/issues/48.

Should this all be done as a pull request for you to later review?
@denar90 do you agree with this?

denar90:
SGTM
just one question about #48 
What exactly do you wanna improve? Passing `flags` from confg file?

pedro93:
I was thinking of populating the pwmetrics object with the properties from the config file in case one exists. If there are repeated parameters both in a config file and from the cmd line, then the cmd line should take precedent.
Imagine something like:

performance.config.js
```js
module.exports = {
    url: "http://github.com",
    json: false,
    runs: 5,
    flags: {
        expectations: true,
    },
    sheets: {
        // sheets configuration
    },
    expectations: {
        ttfmp: {
            warn: '>=3000',
            error: '>=5000',
        },
        tti: {
            warn: '>=5000',
            error: '>=15000',
        },
        ttfcp: {
            warn: '>=1500',
            error: '>=3000',
        },
        psi: {
            warn: '>=3000',
            error: '>=6000',
        },
    },
};
```

Bash
```
$ pwmetrics --config="path/to/file.js" --json
```

The json option in the terminal overrides the config option. A use case for this would be wanting to see results with slightly different options but no having to change the actual file (which you may not have permissions to edit). This would also potentially allow any configuration to be set via the command line that can also be set in the config file.

This logic should be done IMO in the bin/cli.js file after yargs has been merged, hence the dependency.

denar90:
Oh, got your point. Make sense. 
Lets do it 💪  I'm ok to make it in scope of yargs PR. Jut make PR against `yargs` branch, so I'll merge it together.

pedro93:
@paulirish do you approve? If so I will start working on this. I do have a question though, what is the rationale behind this loop?
https://github.com/paulirish/pwmetrics/blob/master/bin/cli.js#L24

denar90:
[...]

PR #51, opened 2017-02-03
Add yargs

denar90:
Resolves #34 

Saving into file will look like

`pwmetrics http://example.com --json pwmetrics-results.json`

instead of 

`pwmetrics http://example.com --json > pwmetrics-results.json`

If `pwmetrics http://example.com --json` then data will be `stdout`

PR #56, opened 2017-02-08
Yargs V2

pedro93:
Yargs re-structure, added the notion of option priority, cleaned pwmetrics constructor to match information in README and added utils object functionality for deep object merging.

I also added the notes specified by @paulirish in https://github.com/paulirish/pwmetrics/pull/51
Fixes: https://github.com/paulirish/pwmetrics/issues/48 and https://github.com/paulirish/pwmetrics/issues/53.

Still need to add tests and fix https://github.com/paulirish/pwmetrics/issues/47, suggestions are welcome on what to do with the data from all runs (save externally?)
@denar90 What do you think?


pedro93:
This is a WIP, @paulirish I noticed when travis failed that the test/fixtures do not match what is in the README about the configurations. Could you please say which one you want to follow, so that I may change the code accordingly or provide an example to follow?

bin/cli.js
```diff
@@ -9,75 +10,96 @@ const yargs = require('yargs');
...
+    'describe': 'Disable CPU throttling',
+    'type': 'boolean',
+    'default': false
+  })
+  .epilogue('For more Lighthouse CLI options see https://github.com/GoogleChrome/lighthouse/#lighthouse-cli-options')
   .check(argv => {
-    if (argv._.length === 0)
-      throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));
+    //test possible url locations, first from cmd line then from config file
+    if (argv._.length !== 0 || (argv.config !== undefined && argv.config.url))
```
denar90: I would like to use `error-first callback` way here
e.g.
```
if (argv._.length === 0 || || (argv.config !== undefined && argv.config.url))
  throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));

return true;
```

bin/cli.js
```diff
@@ -9,75 +10,96 @@ const yargs = require('yargs');
...
-Object.keys(fileConfig).forEach(configPropName => {
-  switch(configPropName) {
-    case 'expectations':
-    case 'sheets':
-      config[configPropName] = fileConfig[configPropName];
-      break;
-  }
-});
+//Merge options from all sources. Order indicates precedence (last => most important)
+const options = mergeDeep(yargs.getOptions().default, cliFlags.config, terminalOptions);
```
denar90: `flags` - it's kinda feature flags for `pwmetrics`, no matter where it passed from, either file or console. Look at the docs - https://github.com/paulirish/pwmetrics#defining-config

`cliFlags.expectations !== config.expectations` 

`typeof cliFlags.expectations === boolean`
`config.expectations === Object`

This is the way I'd like to go.

So going with you suggestion we can simplify this. Just not use `mergeDeep` and use simple `Object.assign`.
If we have flags in console, then we use them, if no - we use config.

```
const config = cliFlags.config;
const url  = cliFlags._[0] || config.url;

Object.assign(config.flags, terminalOptions);

const pwMetrics = new PWMetrics(url, config);
```
smth like this.

bin/cli.js
```diff
@@ -9,75 +10,96 @@ const yargs = require('yargs');
...
+const pwMetrics = new PWMetrics(options.url, options);
 pwMetrics.start()
   .then(data => {
-    data = JSON.stringify(data, null, 2) + '\n';
-    if (cliFlags.json && !cliFlags.json.length) {
-      data && process.stdout.write(data);
-      process.exit(0);
-    } else if (cliFlags.json.length) {
-      write(cliFlags.json, data);
+    if (cliFlags.json) {
```
denar90: `switch` for output type

lib/index.js
```diff
@@ -21,16 +21,17 @@ class PWMetrics {
 
   constructor(url, opts) {
     this.url = url;
-    this.opts = opts;
```
denar90: Feel free to improve tests.

test/config.js
```diff
@@ -0,0 +1,32 @@
+module.exports = {
```
denar90: we already have https://github.com/paulirish/pwmetrics/blob/master/test/fixtures/run-expectations.js

This file is not needed.

bin/cli.js
```diff
@@ -9,75 +10,96 @@ const yargs = require('yargs');
...
-    'runs': 'Does n runs (eg. 3, 5), and reports the median run\'s numbers.',
-    'expectations': 'Expectations from metrics results. Useful for CI. Config required.',
-    'submit': 'Submit metric results into sheets. Config required.',
-    'config': 'Read config form file.',
-    'disable-cpu-throttling': 'Disable CPU throttling',
-    'and more Lighthouse CLI options': 'Docs - https://github.com/GoogleChrome/lighthouse/#lighthouse-cli-options',
+  .wrap(yargs.terminalWidth())
+  .usage('Usage: $0 <url>')
+  .command('url', 'URL to test')
+  .option('json', {
```
denar90: If we are going with output-path we can use output flag instead of json. It might have different values like stdout (default) and json (csv in future). It will add some flexibility.

denar90:
@pedro93 thanks for nice PR 👍

bin/cli.js
pedro93: I was thinking about this earlier today. 

My original idea was to be able in the config file to overwrite any option. Url, flags.expectations, json (the bool) and others, not only the expectations object. 

The reason behind it being easy of use. Simply define all options in a file and be done with it.

Otherwise you have to define expectations (bool) in the bash and the actual expectations values in the file (object).

Thoughts? If it's too radical a change I can revert.

bin/cli.js
pedro93: Sure thing, I can add a set of permitted options and add the necessary functionality.

bin/cli.js
pedro93: Never seen this, how does it work?

test/config.js
pedro93: Ah thanks, I'll remove this then

bin/cli.js
denar90: Ignore his one

bin/cli.js
denar90: It's kinda pattern/way for handling errors. Here some article -http://www.codingdefined.com/2015/10/what-are-error-first-callbacks-in-nodejs.html

bin/cli.js
denar90: It's is right now, you can override flags in config due to doc. I think it's kinda misunderstanding. I used `expectations` as an example 😃 
So in file you can do
```
...
flags: {
  expectations: true,
  submit: true, 
  json: true
}
...
```

You can also ping me on gitter so we can discuss it, if we still not on the same page :)

bin/cli.js
pedro93: Question, when output=json, regardless of output-path, should the graph and expectations be shown?

bin/cli.js
pedro93: I see now what you meant by ```error-first callback```. The reasoning behind this change was that to make the url test work I would have to chain if statements:
```js
    if (argv._.length === 0){
      if(argv.config === undefined) {
        throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));
      } else if(!argv.config.url) {
        throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));
      }
    } 
    return true;
```
Due to possible locations where a url might be found:
 - Console only
 - Config only
 - Console and config

IMO not using the error-first pattern for this particular case is easier to comprehend/reason about, especially if we condense the if statement. 

```js
    if (argv._.length === 0 && (argv.config === undefined || !argv.config.url)){
        throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));
    } 
    return true;
```

bin/cli.js
denar90: Don't show anything, just stdout or write in file.

bin/cli.js
denar90: Yeap, lets use
```
if (argv._.length === 0 && (argv.config === undefined || !argv.config.url)){
        throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));
    } 
    return true;
```

paulirish:
Okay I know this isn't done, but it'll be easier to only have one branch in flight. Merging this into the yargs branch (and the #51 PR)

We can still iterate on things there. Hopefully we can merge to master soon, too. :)

PR #57, opened 2017-02-09
add new image

denar90:
Thoughts?

P.S. Feel free to criticize it!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -6,7 +6,7 @@
 
 CLI tool and lib to gather performance metrics via [Lighthouse](https://github.com/GoogleChrome/lighthouse/). _IN BETA_.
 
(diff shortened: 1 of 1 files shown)


pedro93:
Very nice 👍

paulirish:
Hah! very cool.

I think we should keep the screenshot and a logo separate things.  :)

I do like the general concept of the logo you made here, but we could probably tighten it up, and drop the text. something smaller like 450x230 or whatever.

So let's iterate on the image, and we can refresh the screenshot with a newer version sometime too. :)

denar90:
Thanks guys :)

denar90:
Here is a new one. 
We can also can involve @samccone if he don't mind 😛 

pwmetrics-small

samccone:
I think it is cool ;)

paulirish:
yup. i dig the new one. :)

PR #59, opened 2017-02-10
bring back json flag

paulirish:
@denar90 I think this behavior is what I was trying to explain.

json is a boolean flag. (defaults to false).
output-path handles where things are saved

cc @pedro93

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -18,17 +18,16 @@ const cliFlags = yargs
   .wrap(yargs.terminalWidth())
   .usage('Usage: $0 <url>')
   .command('url', 'URL to test')
-  .option('output', {
-    'describe': 'Specify the output format',
-    'type': 'string',
-    'choices': ['', 'json'],
-    'default': '',
+  .option('json', {
+    'describe': 'Output as json',
+    'type': 'boolean',
+    'default': 'false',
     'group': 'Output:'
   })
   .option('output-path', {
     'describe': 'The file path to output the results',
     'type': 'string',
-    'default': '',
+    'default': 'stdout',
     'group': 'Output:'
   })
   .option('runs', {
@@ -48,7 +47,6 @@ const cliFlags = yargs
   })
   .option('config', {
     'describe': 'Path to json config file',
-    'coerce': (arg) => getConfigFromFile(arg),
     'type': 'string',
   })
   .option('disable-cpu-throttling', {
@@ -65,18 +63,13 @@ const cliFlags = yargs
     return true;
   }).argv;
 
-//Get all explicitly terminal set options, does not include url because url has no default option
-const flags = Object.keys(yargs.getOptions().default).reduce((accum, prop) => {
-  if (cliFlags[prop] !== yargs.getOptions().default[prop])
-    accum[prop] = cliFlags[prop];
-  else if (cliFlags.config && cliFlags.config.flags && cliFlags.config.flags[prop])
-    accum[prop] = cliFlags.config.flags[prop];
-  else accum[prop] = yargs.getOptions().default[prop];
-  return accum;
-}, {});
 
-//Merge options from all sources. Order indicates precedence (last => most important)
-let options = Object.assign({}, cliFlags.config, { flags: flags });
+const config = cliFlags.config ? getConfigFromFile(cliFlags.config) : {};
+
+// For the config flags and URL, we'll allow CLI to override what's set in the config Object
+
+//Merge options from all sources. Order indicates precedence (last one wins)
+let options = Object.assign({}, config, {flags: cliFlags});
 
 //Get url first from cmd line then from config file.
 options.url = cliFlags._[0] || options.url;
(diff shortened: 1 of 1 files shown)


bin/cli.js
```diff
@@ -48,7 +47,6 @@ const cliFlags = yargs
   })
   .option('config', {
     'describe': 'Path to json config file',
-    'coerce': (arg) => getConfigFromFile(arg),
```
paulirish: This feels weird, in that cliFlags contains some preprocessed values. So I moved it to later to do it explicitly.

bin/cli.js
```diff
@@ -65,18 +63,13 @@ const cliFlags = yargs
     return true;
   }).argv;
 
-//Get all explicitly terminal set options, does not include url because url has no default option
```
paulirish: honestly i couldn't figure out why we need this. :)

(i'm mostly going off how the lighthouse CLI works and it's not needed such behavior)

pedro, let me know if i'm missing something.

bin/cli.js
```diff
@@ -65,18 +63,13 @@ const cliFlags = yargs
...
-    accum[prop] = cliFlags.config.flags[prop];
-  else accum[prop] = yargs.getOptions().default[prop];
-  return accum;
-}, {});
 
-//Merge options from all sources. Order indicates precedence (last => most important)
-let options = Object.assign({}, cliFlags.config, { flags: flags });
+const config = cliFlags.config ? getConfigFromFile(cliFlags.config) : {};
+
+// For the config flags and URL, we'll allow CLI to override what's set in the config Object
```
paulirish: this is up for debate, but this precedence order is what is used for the `options` and `options.url` below. happy to change if config needs to trump the CLI.

bin/cli.js
denar90: I think reason doing that is not to rewrite flags in config by default values. Idea about replacing flags in config by passed cli flags requires to do that. 

E.g. if you have flags in config.
```js
{
  {
    json: true,
    expectations: true
  }
}
```

If user didn't pass any cli flags
this `let options = Object.assign({}, config, {flags: cliFlags});` will rewrite config by default values. 

```js
{
  {
    json: false,
    expectations: false
  }
}
```

bin/cli.js
```diff
@@ -65,18 +63,13 @@ const cliFlags = yargs
...
-    accum[prop] = cliFlags[prop];
-  else if (cliFlags.config && cliFlags.config.flags && cliFlags.config.flags[prop])
-    accum[prop] = cliFlags.config.flags[prop];
-  else accum[prop] = yargs.getOptions().default[prop];
-  return accum;
-}, {});
 
-//Merge options from all sources. Order indicates precedence (last => most important)
-let options = Object.assign({}, cliFlags.config, { flags: flags });
+const config = cliFlags.config ? getConfigFromFile(cliFlags.config) : {};
```
denar90: 👍

bin/cli.js
```diff
@@ -95,18 +88,15 @@ const writeToDisk = function (fileName, data) {
...
-      else data && process.stdout.write(data);
+    if (options.flags.json) {
+      // serialize accordingly
+      data = JSON.stringify(data, null, 2) + '\n';
+      // output to file.
+      if (options.flags.outputPath != 'stdout')
+        return writeToDisk(options.flags.outputPath, data);
+      // output to stdout
+      else if (data)
+        process.stdout.write(data);
```
denar90: I think we also need `process.exit(0);` like it was before - 
https://github.com/paulirish/pwmetrics/blob/master/bin/cli.js#L55

bin/cli.js
pedro93: @paulirish, @denar90's train of thought is similar to my own.
My original idea was to be able to unify all options.  Lets say as a user I don't want to type out all explicit flags I need. To solve this I set up a config.file with my intended options.

If for some reason I want to test something out and don't want or can't (current work condition) I want to be able to explicitly overide that specific option in cli.

This approach was the way I found to make this precedence order function without breaking too much:
 - Get everything expliticly set in the command line.
 - Get everything from the config file
 - Whatever has not be yet set is the default value (which yargs defines).

I could have requested a change or submitted a pull request for lighthouse directly but my use case for ```lighthouse``` is merely for performance at the moment and in a visual manner (which ```pwmetrics``` is perfect for). Though to be honest the amount of lighthouse cli options daunts me 😛

bin/cli.js
pedro93: IMO if a user explicitly sets a flag in the CLI I think he/she wants to trump the config file. Otherwise they'll have to remember to check the config file. Thoughts @denar90 ?

bin/cli.js
pedro93: The ```process.exit(0)``` is in the ```then/catch```callbacks. This function is merely for data output processing (serialize and write to stream).

bin/cli.js
pedro93: If we remove this function then we should modify [this line](https://github.com/paulirish/pwmetrics/blob/6f2647fa22c21181e1669ae6e467a845971e94dd/bin/cli.js#L60)  accordingly. When the ```check```function executes it will always (I think) have ```argv.config``` set to empty string which will create false positives.

I'm curious, could you explain a little about your thoughts on this particular line?

denar90: I'm 👍  for keeping value of config as `string`. It's more consistent and architecturaly right. Below we read config from file and mix it with cli options. No magic, very simple to read and operate with config values.

bin/cli.js
denar90: I really don't need this logic right now (mix up config and cli flags) but if @pedro93 is needed, maybe it makes sense to have it.

From another side. We always can add it in for e.g. 2.1 to not block 2.0 release, it's also better to add API then drop it (if drop we will have to make 3.0 release). Maybe after 2.0 more users request this so it will be useful.

bin/cli.js
denar90: Missed second `then`, sorry.

bin/cli.js
pedro93: From my side I can workaround my limitation, so we can leave this as is for now and see if any one else requests the functionality. Either way its not a major change and we can always refer back to this pull request to see how to apply it.

bin/cli.js
pedro93: I was merely trying to take advantage of the ```yargs``` library as much as possible, not particularly critical IMO.

PR #50, opened 2017-02-03
Use lighthouse metricsDefinitions

denar90:
Try [Metrics.metricsDefinitions](https://github.com/GoogleChrome/lighthouse/blob/ee8d967cabc5d88a097c71d91c83e0bb161c15fe/lighthouse-core/lib/traces/pwmetrics-events.js#L32) API

PR #60, opened 2017-02-17
add start typescript setup

denar90:
- typescript configurations
- sheets.js -> sheets.ts

### DIFF
--- a/types/types.ts
+++ b/types/types.ts
@@ -18,14 +18,15 @@ interface MetricsResults {
 
 interface Timestamps {
   title: string;
-  name: string;
-  value: number;
+  id: string;
+  timestamp: number;
 }
 
 interface Timings {
   title: string,
-  name: string;
-  value: number;
+  id: string;
+  timestamp: number;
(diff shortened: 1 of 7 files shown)


pedro93:
Ping @paulirish, any objection to this?

types/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  type: string;
+  options: {
+    spreadsheetId: string;
+    tableName: string;
+    clientSecret: Object;
+  };
+}
+
+interface MetricsResults {
+  timestamps: Array<any>;
```
samccone: would prefer if we defined the type instead of any

denar90: Good point. I was thinking to do that after merging https://github.com/paulirish/pwmetrics/pull/50

denar90:
@samccone thanks for the review 👍  I'm merging this one, if you have any comments for improving this stuff I can make it in another PR.

paulirish:
Lgtm!

PR #62, opened 2017-02-25
cleanup

paulirish:
A few things.

1. Based on the readme, any settings in package.json are read without even indicating `--config`. Do we want this? If so, I changed it to follow that style. It works for me.

2. nuked a stray debugger statement.

3. minor tweaks to the sheets setup docs.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -57,7 +57,7 @@ const cliFlags = yargs
   .epilogue('For more Lighthouse CLI options see https://github.com/GoogleChrome/lighthouse/#lighthouse-cli-options')
   .argv;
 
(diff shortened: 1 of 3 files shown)


denar90:
>Based on the readme, any settings in package.json are read without even indicating --config. Do we want this?

Wanna remove it?
The idea was to reduce number of files in repo which will use `pwmetrics`. If you have simple site and wanna just check expectations during build - config in `package.json` is perfect. In other case, if you use gsheets with keys and stuff, it will be file.

paulirish:
yup. sgtm.

so config from package.json is a default.  --config arg is only used to specify another file which has the deets.

Issue #65, opened 2017-02-27
Cleaner error messaging when using invalid parameters


pedro93:
Just a note, when running the cli tool with invalid parameters a cleaner output should be shown:

![screen shot 2017-02-27 at 12 29 14](https://cloud.githubusercontent.com/assets/3718654/23361432/8c4849dc-fce8-11e6-867d-0b91ccebf4ca.png)

This happens because we have no try catch block to handle this exception. We can fix this by doing the following In the [cli.js file test section](https://github.com/paulirish/pwmetrics/blob/master/bin/cli.js#L68)
```js
if (!options.url || !options.url.length) {
   console.err(getMessage('NO_URL'));
   process.exit(-1);
}
```

Also I would prefer the Error message outputed from getMessageWithPrefix, its a little more visual for the user.

denar90:
solved by #67

Issue #43, opened 2017-01-29
Add eslint


denar90:

pedro93:
I'm 100% in favor of this. @paulirish @denar90 do either of you have any specific eslint rule files you want to use or shall we use [lighthouse's one](https://github.com/GoogleChrome/lighthouse/blob/master/.eslintrc.js) ?

paulirish:
Id be happy to use lighthouses rules

pedro93:
Since we added and merged eslint rules in https://github.com/paulirish/pwmetrics/pull/73 im closing this, feel free to reopen.

PR #67, opened 2017-03-03
WIP: Added no url error and —help message if cli tool misused.

pedro93:
This is a __work in progress__, some code cleanup currently on cli.
Its very similar to the Yargs V2 merge request but for project-consistency offers functionality akin to lighthouse at the cli-level. 

__Edit__: The tests fail because of https://github.com/paulirish/pwmetrics/blob/master/test/cli.js. Does it make sense for a cli tool to throw an Error that the OS/terminal process will catch?

__Edit 2__: Addresses https://github.com/paulirish/pwmetrics/issues/65 :

screen shot 2017-03-07 at 09 17 08

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -10,6 +10,7 @@ const yargs = require('yargs');
 const PWMetrics = require('../lib/index');
 const { getConfigFromFile } = require('../lib/utils/fs');
 const { getMessageWithPrefix, getMessage } = require('../lib/utils/messages');
+let config;
 
 const cliFlags = yargs
   .help('help')
@@ -47,9 +48,7 @@ const cliFlags = yargs
   })
   .option('config', {
     'describe': 'Path to config file',
-    'type': 'string',
-    //equivalent to: cliFlags.config = getConfigFromFile(cliFlags.config);
-    'coerce': getConfigFromFile 
+    'type': 'string'    
   })
   .option('disable-cpu-throttling', {
     'describe': 'Disable CPU throttling',
@@ -58,7 +57,9 @@ const cliFlags = yargs
   })
   .check((argv) => {
     // Make sure pwmetrics has been passed a url, either from cli or config file
-    if (argv._.length === 0 && (argv.config === undefined || !argv.config.url))
+    if(argv.config) config = getConfigFromFile(argv.config);
+
+    if (argv._.length === 0 && (config === undefined || !config.url))
         throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));
 
     return true;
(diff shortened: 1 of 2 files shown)


bin/cli.js
```diff
@@ -48,26 +48,30 @@ const cliFlags = yargs
   .option('config', {
     'describe': 'Path to config file',
     'type': 'string',
+    //equivalent to: cliFlags.config = getConfigFromFile(cliFlags.config);
+    'coerce': getConfigFromFile
```
denar90: I though we agreed to leave it as string (path to file) instead of mixing up path and config value...
Am I wrong?

pedro93: You are not. I realized that in order to have the "--help" message like lighthouse does then yargs must be able to validate if the url is being supplied anywhere. If the argv object in the check function throws an error (which is the red no url message) it will also show the see help for more message. In order to do so it must check if the url is given by the cli or config object which during that check is only a link. It must have that object data beforehand, hence the coerce. From a users point of view it's still a link, this coerce only takes that argument and attempts to read its content. It's the exact same functionality as the code it replaces

bin/cli.js
```diff
@@ -47,27 +48,31 @@ const cliFlags = yargs
...
     'describe': 'Path to config file',
-    'type': 'string',
+    'type': 'string'    
   })
   .option('disable-cpu-throttling', {
     'describe': 'Disable CPU throttling',
     'type': 'boolean',
     'default': false
   })
+  .check((argv) => {
```
denar90: Nice 👍 
Just one nit pick

What do you think if we do smth like
```
....
const url = argv._ || config.url;
if (!url.length)
  throw new Error(getMessageWithPrefix('ERROR', 'NO_URL'));

return true;
```
?

pedro93: Inside the check function? Perfectly fine by me, i'll make the modification

denar90: LGTM
I thought `getConfigFromFile ` returns empty object if there no config data though...

bin/cli.js
```diff
@@ -47,20 +48,27 @@ const cliFlags = yargs
   })
   .option('config', {
     'describe': 'Path to config file',
-    'type': 'string',
+    'type': 'string'
```
paulirish: trailing ws

bin/cli.js
```diff
@@ -47,20 +48,27 @@ const cliFlags = yargs
...
+    'type': 'string'    
   })
   .option('disable-cpu-throttling', {
     'describe': 'Disable CPU throttling',
     'type': 'boolean',
     'default': false
   })
+  .check((argv) => {
+    // Make sure pwmetrics has been passed a url, either from cli or config file
+    if(argv.config) config = getConfigFromFile(argv.config);
```
paulirish: add space before paren

pedro93: Add eslint to catch these things https://github.com/paulirish/pwmetrics/issues/43 ? 😛

pedro93:
I think this is know ready to merge if approved

PR #69, opened 2017-03-06
update lighthouse version, improve build

denar90:
- update lighthouse version
 - move recipes run into separate build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ matrix:
   include:
     - node_js: "6"
     - node_js: "7"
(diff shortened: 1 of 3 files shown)


paulirish:
why the move to do recipes in a separate build? faster?

denar90:
Recipes kinda integration tests, they are always failing (I hope it will be not ages) and confusing (is it unit test fail or recipes). Unit tests pass fine, so I thought why not to move recipes to separate build.

paulirish:
sg

Issue #64, opened 2017-02-26
Add a more descriptive output for the runs. 


rpkoller:
At the moment if you are running an audit with more than one run you only get the generic line `Launching Chrome`: 

![bildschirmfoto 2017-02-26 um 14 49 01](https://cloud.githubusercontent.com/assets/1665422/23340366/1eeb465e-fc35-11e6-9bf8-c9b9977c228f.png)

In my opinion it would be more descriptive and helpful if the output would say something like `Run 1 out of 5 is finished successfully` and so on. That way you would also have the opportunity to deal with certain exceptions if something went wrong within the microcopy like `Unable to complete run 3 due to .....`

pedro93:
@rpkoller @denar90 @paulirish if a run fails should it be re-run to ensure at least n runs successfully occur if a user specifies an average?

rpkoller:
@pedro93 I would vote against the `Launching Chrome`message. That is the initial key of my suggestion to replace it with a more meaningful line like `Run 1 out of 5 is finished successfully`, so the user is aware that the first round of testing was finished successfully. `Launching Chrome`is a bit abstract and you see that anyway when the app focus switches from the terminal to Chrome. Therefor the `Run 1 out of 5 is finished successfully` is more reassuring. And in the case of an error you might introduce a default behaviour either just note it or rerun the test like you suggested. e.g. in case reruns are default you might write `pwmetrics http://example.com/ --runs=3 --no-rerun` as an override and you just get `Run 3 failed due to exception xyz`

pedro93:
@paulirish @denar90 what do you think. I believe this is possible to do in the scope of the existing pull request. The re-run functionality I have to think about how to implement though given the current code structure.

pedro93:
@rpkoller I understand your point of view but what about running pwmetrics in a CI system? It is helpful to know what stage of the process pwmetrics is in. I concede though that a static message does not give a lot of information

rpkoller:
uh i haven't beared the CI case in mind. then you are totally right! sorry... But a suggestion for the output anyway: 

```
Run 1: Launching Chrome
Run 1: Run 1 of 3 finished successfully
```
That way it would be more aligned layout wise. and you could places the checkmark and the icon for failed processes easily prepended. and i would maybe add an empty line before the first `launching chrome` line and between the last run line and the  `*median run *` line so it looks a bit more cleaned up, loose and not that packed.

pedro93:
[...]

PR #70, opened 2017-03-07
WIP: Added a more descriptive output for the runs.

pedro93:
Addresses https://github.com/paulirish/pwmetrics/issues/64. Very minor code change.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -53,10 +53,10 @@ class PWMetrics {
     const allDone = results.reduce((chain, run, index) => {
       return chain.then(_ => {
         return this.run().then(data => {
-          console.log("Run",index+1 , "of", results.length, "finished successfully");
+          console.log(getMessageWithPrefix('SUCCESS', 'SUCCESS_RUN', index, results.length));
           results[index] = data;
         }).catch(error => {
-          console.log("unable to complete run", index+1, "of", results.length, "due to", error.message);
+          console.error(getMessageWithPrefix('ERROR', 'FAILED_RUN', index, results.length, error.message));
           throw new Error(error);
         });
       });
@@ -142,7 +142,7 @@ class PWMetrics {
       }
 
       return Promise.resolve(preparedData);
-    } catch(error) {
+    } catch (error) {
       return Promise.reject(error);
     }
   }
--- a/lib/index.js
+++ b/lib/index.js
@@ -53,7 +53,11 @@ class PWMetrics {
     const allDone = results.reduce((chain, run, index) => {
       return chain.then(_ => {
         return this.run().then(data => {
+          console.log("Run",index+1 , "of", results.length, "finished successfully");
           results[index] = data;
+        }).catch(error => {
+          console.log("unable to complete run", index+1, "of", results.length, "due to", error.message);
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -53,7 +53,11 @@ class PWMetrics {
     const allDone = results.reduce((chain, run, index) => {
       return chain.then(_ => {
         return this.run().then(data => {
+          console.log("Run",index+1 , "of", results.length, "finished successfully");
           results[index] = data;
+        }).catch(error => {
+          console.log("unable to complete run", index+1, "of", results.length, "due to", error.message);
```
paulirish: can you change to use template strings?

denar90: + can you move text into messages and use `getMessage` for this?

pedro93:
Here is the updated output, @rpkoller is this better for you?

screen shot 2017-03-08 at 12 36 32

lib/index.js
pedro93: Done both, please see new commit and program output in the screenshot I commented :)

rpkoller:
Looks great! The only thing i am still unsure if `Launching Chrome`is necessary at all since you see the opening chrome window anyway.

lib/utils/messages.js
```diff
@@ -13,56 +13,60 @@ const redify = str => `${RED}${str}${RESET}`;
...
+    case 'vc':
+      return 'Visually Complete 100%';
+    case 'tti':
+      return 'Time to Interactive';
+    case 'vc85':
+      return 'Visually Complete 85%';
+    case 'SUCCESS_RUN':
+      return `Run ${args[0][0] + 1} of ${args[0][1]} finished successfully`;
+    case 'FAILED_RUN':
+      return `Unable to complete run ${args[0][0] + 1} of ${args[0][1]} due to ${args[0][2]}`
```
pedro93: Never worked with rest parameters before, this was the way I found to access the variable data needed. Is it correct?

denar90: Just do smth like
`Run ${args[0] + 1} of ${args[1]} finished successfully`

denar90:
We are launching/killing chrome for each run. It's useful to have this message when errors happens or for CI as @pedro93 noticed.

lib/utils/messages.js
pedro93: Args[0] is an array like this: [index, results.length] e.g: [1,3] with args being [[1,3]], not sure why though. I have a suspicion it's the nested rest parameters generated by getmessagewithprefix

pedro93: Here is quick (it's bad don't judge) code snippet demonstrating what was happening: 
https://repl.it/GNlK/0

 I fixed it in this [commit](https://github.com/paulirish/pwmetrics/pull/70/commits/f2bd2e1afa5f697fced9d5e5c3c934bc6de1b809). Doesn't seem to break anything, can anyone confirm @paulirish @denar90 ?

denar90:
@paulirish can we merge this one?

pedro93:
@denar90 @paulirish with your permission I would merge both this PR and the eslint one tomorrow.

Issue #52, opened 2017-02-06
Add some of features from pwmetrics-helper


denar90:
Right now `pwmetrics` can do more then [pwmetrics-helper](https://github.com/addyosmani/pwmetrics-helper), to have one place to manage progressive web metrics I propose to add some missed features  (get `desiredMetric `) which `pwmetrics-helper` has and propose @addyosmani to deprecate it.

pedro93:
I agree especially after https://github.com/paulirish/pwmetrics/pull/51 merges.

denar90:
So we should just add `desiredMetric` feature.
Should we draw chart?
Should we check expectations if `expectations` flag and `desiredMetric` are set? (personally 👍)

pedro93:
Just to bring this alive, should we merely add the code from that repo or re-create the functionality in a more integrated manner?  @denar90 @paulirish depending on how strongly you feel, it would require a large code re-write.

denar90:
The code which can be reused is just one function :)
https://github.com/addyosmani/pwmetrics-helper/blob/master/index.js#L30-L32

You can easy copy/paste it, add cli flags and stuff.

pedro93:
Does it make sense to have a cli flag for this? if so, how should it look? This functionality is is more of an API addition.

denar90:
I think `desiredMetric` will be fine. It should be possible to use through cli and API.

pedro93:
Should desiredMetric be applied to the median run or all runs, should you only specify one or can you provide a list to run? Also we can do an improvement here that is to tell lighthouse to run the specified audits. As of right now and https://github.com/addyosmani/pwmetrics-helper does this, is to run lighthouse as is right now (with perf audits) and filter the results for only the desired metric (only 1).

PR #71, opened 2017-03-07
DesiredMetric functionality.

pedro93:
See https://github.com/paulirish/pwmetrics/issues/52 for some context on possible pull request improvements and current state.

Any feedback is welcome. 

On a more personal side note, I would suggest the following additional modifications:
 - Refractor the metrics file into typescript with this pull request.
 - With this new class, add metric validation both for the user selected metrics and data received from lighthouse.
 - Allow the possibility to add custom performance-related audits (assumes audit generates a report based on pre-defined schema). Not sure how to do it, don't have enough knowledge to create custom audits.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -57,7 +57,7 @@ const cliFlags = yargs
   .option('desiredMetrics', {
     'describe': 'Array of metrics to be obtained',
     'type': 'array',
-    'deafult': false
+    'default': false
   })
   .epilogue('For more Lighthouse CLI options see https://github.com/GoogleChrome/lighthouse/#lighthouse-cli-options')
   .argv;
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -57,8 +57,7 @@ const cliFlags = yargs
   })
   .option('desiredMetrics', {
     'describe': 'Array of metrics to be obtained',
-    'type': 'array',
-    'default': false
+    'type': 'array'
   })
   .check( (argv) => {
     // Make sure pwmetrics has been passed a url, either from cli or config file
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -54,7 +54,7 @@ const cliFlags = yargs
     'type': 'boolean',
     'default': false
   })
-  .option('desiredMetric', {
+  .option('desiredMetrics', {
     'describe': 'Array of metrics to be obtained',
     'type': 'array',
     'deafult': false
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -54,6 +54,11 @@ const cliFlags = yargs
     'type': 'boolean',
     'default': false
   })
+  .option('desiredMetric', {
+    'describe': 'Array of metrics to be obtained',
+    'type': 'array',
+    'deafult': false
+  })
   .epilogue('For more Lighthouse CLI options see https://github.com/GoogleChrome/lighthouse/#lighthouse-cli-options')
   .argv;
 
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -59,6 +59,7 @@ const cliFlags = yargs
     'describe': 'Array of metrics to be obtained',
     'type': 'array',
     'default': false
+  })
   .check( (argv) => {
     // Make sure pwmetrics has been passed a url, either from cli or config file
     if(argv.config) config = getConfigFromFile(argv.config);
--- a/lib/index.js
+++ b/lib/index.js
@@ -30,7 +30,7 @@ class PWMetrics {
     this.sheets = opts.sheets;
     this.expectations = opts.expectations || false;
     this.tryLighthouseCounter = 0;
-    this.desiredMetrics = opts.flags.desiredMetric || [];
+    this.desiredMetrics = this.flags.desiredMetrics || [];
 
     if (this.flags.expectations) {
       if (this.expectations) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -27,7 +27,7 @@ class PWMetrics {
     this.sheets = opts.sheets;
     this.expectations = opts.expectations || false;
     this.tryLighthouseCounter = 0;
-    this.desiredMetrics = this.flags.desiredMetrics || [];
+    this.desiredMetrics = this.flags.desiredMetrics || ['tti', 'psi', 'ttfmp', 'ttfcp'];
 
     if (this.flags.expectations) {
       if (this.expectations) {
@@ -55,7 +55,8 @@ class PWMetrics {
           results[index] = data;
         }).catch(error => {
           console.error(getMessageWithPrefix('ERROR', 'FAILED_RUN', index, results.length, error.message));
-          throw new Error(error);
+          console.error(error.stack);
+          Promise.reject(error);
         });
       });
     }, Promise.resolve());
@@ -63,12 +64,14 @@ class PWMetrics {
     return allDone.then(_ => {
       const ret = {runs: results};
       if (this.runs > 1 && !this.flags.submit) {
-        ret.median = this.findMedianRun(results);
         console.log(getMessage('MEDIAN_RUN'));
+        ret.median = this.findMedianRun(results);
         this.displayOutput(ret.median);
       } else if (this.flags.submit) {
         const sheets = new Sheets(this.sheets);
         return sheets.appendResults(ret.runs).then(_ => ret);
+      } else {
+        this.displayOutput(ret.runs[0]);
       }
       return ret;
     });
@@ -93,7 +96,7 @@ class PWMetrics {
         return this.launcher.kill().then(_ => data);
       })
       .catch(error => {
-        throw new Error(error);
+        return Promise.reject(error);
       });
   }
 
@@ -130,10 +133,6 @@ class PWMetrics {
   recordLighthouseTrace(data) {
     try {
       const preparedData = metrics.prepareData(data, this.desiredMetrics);
-
-      if (!this.flags.submit && this.runs <= 1) {
-        this.displayOutput(preparedData);
-      }
 
       if (this.flags.expectations) {
         expectations.checkExpectations(preparedData.timings, this.expectations);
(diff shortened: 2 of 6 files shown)


paulirish:
I was late to weigh in on #52 but I'm not convinced this is totally necessary. This is pretty much a matter of filtering the results, which should be pretty straightforward for any consumer of pwmetrics.

I don't think we necessarily need to add this feature. 

It would be more neccessary if Lighthouse was doing scrolling/animation/input perf tests too and you wanted results faster without checking everyything..

pedro93:
Im not near my computer to show but, imagine a scenario where a user doesn't want to see the graph for the default metrics but for ones he/she specifies. With this pull request that becomes possible.

This is something I have a use for and would use right away :)

bin/cli.js
```diff
@@ -54,6 +54,11 @@ const cliFlags = yargs
     'type': 'boolean',
     'default': false
   })
+  .option('desiredMetrics', {
+    'describe': 'Array of metrics to be obtained',
+    'type': 'array',
+    'deafult': false
```
paulirish: default

lib/metrics.js
```diff
@@ -49,7 +49,16 @@ function prepareData(res) {
     timestamp: navStart.getTs(audits)
   }];
 
-  metricsDefinitions
+
+  let filteredDefinitions = metricsDefinitions;
+
+  if (desiredMetrics.length > 0) {
+    filteredDefinitions = metricsDefinitions.filter((def) => {
+      return desiredMetrics.includes(def.id);
```
paulirish: might as well inline this

denar90:
It was my idea to bring it here. The main point of it, just make ecosystem more clear deprecating pwmetrics-helper.

lib/metrics.js
```diff
@@ -49,7 +49,16 @@ function prepareData(res) {
     timestamp: navStart.getTs(audits)
   }];
 
-  metricsDefinitions
+
+  let filteredDefinitions = metricsDefinitions;
+
+  if (desiredMetrics.length > 0) {
```
denar90: > tip: you can use just `if (desiredMetrics.length)` :)

pedro93:
Here is a screenshot of how to use the functionality. Does it make sense?

screen shot 2017-03-08 at 10 27 09

pedro93:
This merge request has a corner case where if no ```TTI```  audit is requested then it crashes on doing the median run calculations, see https://github.com/paulirish/pwmetrics/issues/72 for more

paulirish:
will comment on #72..

test/index.js
```diff
@@ -74,7 +74,6 @@ describe('PWMetrics', () => {
       it('should not call methods for calculating median results', () => {
         return pwMetrics.start().then(_ => {
           expect(findMedianRunStub).to.not.have.been.called;
-          expect(displayOutputStub).to.not.have.been.called;
```
pedro93: I did this because when only 1 run is specfied the code should display the graph. 
This change matches that behaviour, see https://github.com/paulirish/pwmetrics/pull/71/files#diff-6d186b954a58d5bb740f73d84fe39073L133

pedro93:
Build failed due to: https://github.com/yarnpkg/yarn/issues/2915 Possible left-pad-like issues?

lib/index.js
```diff
@@ -27,6 +27,17 @@ class PWMetrics {
     this.sheets = opts.sheets;
     this.expectations = opts.expectations || false;
     this.tryLighthouseCounter = 0;
+    this.desiredMetrics = [];
```
denar90: you can throw it.

lib/index.js
```diff
@@ -27,6 +27,17 @@ class PWMetrics {
...
+    if (this.flags.desiredMetrics) {
+      this.desiredMetrics = this.flags.desiredMetrics.filter(metric => {
+        if (!Object.values(metrics.ids).includes(metric)) {
+          console.log(getMessageWithPrefix('WARNING', 'UNKOWN_METRIC', metric));
+          return false;
+        }
+        return true;
+      });
+    }
+    this.desiredMetrics = this.desiredMetrics.length ? this.desiredMetrics : ['tti', 'psi', 'ttfmp', 'ttfcp'];
```
denar90: easier just to do smth `this.desiredMetrics = Object.assign([], this.flags.desiredMetrics, ['tti', 'psi', 'ttfmp', 'ttfcp']);

Also better to move `['tti', 'psi', 'ttfmp', 'ttfcp']` into const in scope of `metrics.js` and require it in this file.

lib/index.js
```diff
@@ -27,6 +27,17 @@ class PWMetrics {
     this.sheets = opts.sheets;
     this.expectations = opts.expectations || false;
     this.tryLighthouseCounter = 0;
+    this.desiredMetrics = [];
+    if (this.flags.desiredMetrics) {
+      this.desiredMetrics = this.flags.desiredMetrics.filter(metric => {
```
denar90: `this.flags.desiredMetrics = this.flags.desiredMetrics...`

lib/index.js
```diff
@@ -54,20 +65,23 @@ class PWMetrics {
           results[index] = data;
         }).catch(error => {
           console.error(getMessageWithPrefix('ERROR', 'FAILED_RUN', index, results.length, error.message));
-          throw new Error(error);
+          console.error(error.stack + '\n');
```
denar90: Why don't we wanna throw error here?

lib/index.js
```diff
@@ -54,20 +65,23 @@ class PWMetrics {
...
         });
       });
     }, Promise.resolve());
 
     return allDone.then(_ => {
-      const ret = {runs: results};
-      if (this.runs > 1 && !this.flags.submit) {
-        ret.median = this.findMedianRun(results);
+      const ret = {runs: results.filter(r => !(r instanceof Error))};
+      if (this.runs > 1 && !this.flags.submit && ret.runs.length) {
```
denar90: Why do we need `ret.runs` ?

lib/metrics.js
```diff
@@ -32,7 +29,7 @@ const checkAudits = audits => Object.keys(audits).forEach(key => {
     throw new Error(`${debugString} Audit key: ${key}`);
 });
 
-function prepareData(res) {
+function prepareData(res, desiredMetrics = []) {
```
denar90: It's not possible to have `desiredMetrics ` empty array due to https://github.com/paulirish/pwmetrics/pull/71/files#diff-6d186b954a58d5bb740f73d84fe39073R40

lib/metrics.js
```diff
@@ -49,7 +46,13 @@ function prepareData(res) {
     timestamp: navStart.getTs(audits)
   }];
 
-  metricsDefinitions
+
+  let filteredDefinitions = metricsDefinitions;
+
+  if (desiredMetrics.length)
```
denar90: drop it

lib/index.js
pedro93: We already have it, I merely don't include the failed runs so it doesn't crash the findMedian function

lib/index.js
pedro93: imagine you don't want the TTI metric because it's unstable or doesn't bring you wanted data. If that's the case, object assign will not work. You start with the defaults and add to the default array, no way to remove from default.

lib/index.js
pedro93: What you get from input may be different from what pwmetrics can give you. This code validates that. You tell it to run metric XYZ, the desiredMetric flag has content but it is not valid-> remove from array.

lib/index.js
denar90: Yeap, sorry, vice versa``this.desiredMetrics = Object.assign([], ['tti', 'psi', 'ttfmp', 'ttfcp'], this.flags.desiredMetrics);`

pedro93: Consider this code snippet:
```js
var metrics = Object.assign([], ['tti', 'psi', 'ttfmp', 'ttfcp'], ['fv']);
// metrics = [ 'fv', 'psi', 'ttfmp', 'ttfcp' ]
```
 The user only wants the fv metric, if we use that code they'll get unwanted metrics

pedro93: Because it would send an error to https://github.com/pedro93/pwmetrics/blob/145b489f98636a5d37c7f2e5fe8523deed965f38/lib/index.js#L74 which does not have a catch clause. Suppose the run 4/10 throws, by throwing the error, the remaining runs are not completed:

With the throw Error I had to manually cancel the process (gave it 5m to recover).
screen shot 2017-03-16 at 09 19 25

lib/index.js
pedro93: Because it would send an error to https://github.com/pedro93/pwmetrics/blob/145b489f98636a5d37c7f2e5fe8523deed965f38/lib/index.js#L74 which does not have a catch clause. Suppose the run 4/10 throws, by throwing the error, the remaining runs are not completed:

With the throw Error I had to manually cancel the process (gave it 5m to recover).
screen shot 2017-03-16 at 09 19 25

lib/index.js
pedro93: This is here to ensure that the .length property of [this.desiredMetrics](https://github.com/pedro93/pwmetrics/blob/145b489f98636a5d37c7f2e5fe8523deed965f38/lib/index.js#L40) does not crash. Imagine you are running pwmetrics as a node module and don't pass the desiredMetrics, the code must have a fallback.

lib/index.js
denar90: But using `console.error(error.stack + '\n');` you output the trace and it feels like error throw, which is not. If you wanna show error and continue process, then what about consoling just error message without trace to avoid confusing.

lib/metrics.js
```diff
@@ -17,9 +17,6 @@ const metricsIds = {
 
 module.exports = {
   hiddenMetrics: [
-    metricsIds.FV,
-    metricsIds.VC100,
-    metricsIds.VC85,
```
denar90: We should have them. Because we don't wanna show them on chart but show in file output.

lib/index.js
```diff
@@ -198,14 +208,29 @@ class PWMetrics {
   }
 
   findMedianRun(results) {
-    const ttiValues = results.map(r => r.timings.find(timing => timing.id === metrics.ids.TTI).timing);
-    const medianTTI = median(ttiValues);
+    const maxTimingValue = Math.max(...results.reduce((acc, r) => {
+      // Result may have invalid data -> failed run
+      if (r.timings)
```
denar90: if run was failed should it be present in results at all?

pedro93: You are right this redundant, the error filter in the AllDone.then function should deal with this.

lib/metrics.js
pedro93: Can we not give that freedom to the user? Let them say which metrics they want to see, both in the chart and output files?

lib/index.js
pedro93: I put trace to help identify where the error came from. It has happened to me that an error was thrown not because of Lighthouse but because of something else, this trace allows users to understand that. It's also my own interpretation and improvement of this: https://github.com/paulirish/pwmetrics/issues/64#issuecomment-285015545. It's a minor change and I don't feel particularly attached to it.

lib/metrics.js
denar90: Yeah but by default I wanna see `['tti', 'psi', 'ttfmp', 'ttfcp']` in chart and all bunch of metrics in file...

pedro93: Ok, would you agree if a user specifies a metric it should be shown in the graph but if not it should revert to showing the defaults? Any suggestions?

denar90: yeap, smth like this

pedro93:
After talking with @denar90 I am closing this PR. As it is, it involves many changes to the codebase some of which are not trivial and the resulting output is not worth the change. If anyone wants to look at the code, it will be in my fork of the project.

PR #73, opened 2017-03-08
Add eslint rules…

pedro93:
To have some linting rules in the repository, response to: https://github.com/paulirish/pwmetrics/issues/43

125 problems to be solved (no-undef, rest params and max-length errors) need help...

@denar90 @paulirish This is not yet ready but if either of you know how to solve the remaining problems I would love some help, this is beyond my knowledge 😝 . Perhaps we can merge to master and we finish it together?

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,13 @@
+# Ignore js files inside local chromium download on Travis.
+**/chrome-linux/**
+**/chrome-win32/**
+
+**/node_modules/**
+**/third_party/**
+
+**/closure/*/*
+coverage/**
+
+# Typescript compiled
+lib/sheets.js
+types/*.js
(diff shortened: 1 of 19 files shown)


denar90:
Since we have a lot of eslint errors I have an idea. What if we merge eslint and stuff around it with commented rules for eslint config in scope of this pr?
Then step by step, uncommenting rule by rule fixing issues.
Profit from this approach: reduced number of possible merge conflicts, easier to fix them (easier to fix 10 small issue the 1 big :) )

pedro93:
I like it :) I can fix the merge conflicts and comment out the rules. Then we merge it and fix incrementally.

paulirish:
does `--fix` get us very far?  but yeah agree that right now it's a little tricky because conflicts.

we could set most of the errors to warnings and land it like that.

pedro93:
I didn't know about that flag, I manually went into each file and used vscode's integration with eslint, it didn't solve those though. The PR already sets the warning. May I suggest we merge it and fix the warnings post-PR?

paulirish:
Sgtm

PR #78, opened 2017-03-21
metrics.js -> metrics.ts

denar90:
@samccone if you don't mind. 
This is small one :)

🍰 this time 😛

### DIFF
--- a/lib/metrics.ts
+++ b/lib/metrics.ts
@@ -1,8 +1,9 @@
 // Copyright 2016 Google Inc. All Rights Reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE
-'use strict';
 
 const metricsDefinitions = require('lighthouse/lighthouse-core/lib/traces/pwmetrics-events.js').metricsDefinitions;
+
+import {MetricsResults, MetricsDefinitions, Timings, Timestamps, LighthouseResults, LighthouseAudits} from '../types/types';
 
 const metricsIds = {
   NAVSTART: 'navstart',
@@ -26,14 +27,14 @@ module.exports = {
   prepareData
 };
 
-const checkAudits = audits => Object.keys(audits).forEach(key => {
+const checkAudits = (audits: LighthouseAudits) => Object.keys(audits).forEach(key => {
   const debugString = audits[key].debugString;
   if (audits[key].debugString)
     throw new Error(`${debugString} Audit key: ${key}`);
 });
 
-function prepareData(res) {
-  const audits = res.audits;
+function prepareData(res: LighthouseResults): MetricsResults {
+  const audits: LighthouseAudits = res.audits;
 
   checkAudits(audits);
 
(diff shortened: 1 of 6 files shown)


denar90:
@pedro93 if you are ok with this one => 💥 

@samccone if here smth to improve I'll make another PR then.
👯 🍺  from me :)

lib/metrics.ts
```diff
@@ -26,50 +27,45 @@ module.exports = {
...
   checkAudits(audits);
 
   const colorP0 = 'yellow';
   const colorP2 = 'green';
   const colorVisual = 'blue';
 
-  const timings = [];
-  const navStart = metricsDefinitions.find(def => def.id === metricsIds.NAVSTART);
-  const timestamps = [{
+  const timings: Array<Timings> = [];
```
samccone: for simple array types prefer Timings[]

lib/metrics.ts
```diff
@@ -26,50 +27,45 @@ module.exports = {
...
+  const navStart = metricsDefinitions.find((def: MetricsDefinitions) => def.id === metricsIds.NAVSTART);
```
samccone: def: type should be inferred

lib/metrics.ts
```diff
@@ -26,50 +27,45 @@ module.exports = {
...
     id: navStart.id,
     timestamp: navStart.getTs(audits)
   }];
 
   metricsDefinitions
-    .filter(def => def.id !== metricsIds.NAVSTART)
-    .forEach(metric => {
-      const resolvedMetric = {
+    .filter((def: MetricsDefinitions) => def.id !== metricsIds.NAVSTART)
+    .forEach((metric: MetricsDefinitions) => {
```
samccone: again, should not be needed here :)

types/types.ts
```diff
@@ -1,6 +1,19 @@
 // Copyright 2016 Google Inc. All Rights Reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE
 
+//@todo remove stuff after https://github.com/GoogleChrome/lighthouse/pull/1882 be merged and version released
```
samccone: looks like this has landed

types/types.ts
```diff
@@ -19,6 +32,12 @@ interface MetricsResults {
   initialUrl: string;
 }
 
+interface MetricsDefinitions {
+  name: string;
+  id: string;
+  getTs(audits: any): number;
```
samccone: is this really an unknown type? or can we type the any :)

types/types.ts
pedro93: Has it been released to npm or is it in the master branch?

denar90: Yeap, we should wait a bit for 1.6.4, 1.6.3 doesn't have it...

lib/metrics.ts
```diff
@@ -26,50 +28,45 @@ module.exports = {
...
+const checkAudits = (audits: LighthouseAudits) => Object.keys(audits).forEach(key => {
   const debugString = audits[key].debugString;
   if (audits[key].debugString)
     throw new Error(`${debugString} Audit key: ${key}`);
 });
 
-function prepareData(res) {
-  const audits = res.audits;
+function prepareData(res: LighthouseResults): MetricsResults {
+  const audits: LighthouseAudits = res.audits;
```
samccone: this should be inferred

paulirish:
sg. lets land after the nit is addressed.

denar90:
@samccone @paulirish thanks 🎉 
I think one last big refactoring PR will be landed for `index.js` in a few days. Count on your review if you don't mind :)

Issue #72, opened 2017-03-08
Pwmetrics fails with pair number of runs


pedro93:
When asking for an average of n runs (where n is a pair) the program will crash because of the calculations done in the [findMedianRun function](https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L204). 

Specifically what is happening is that the program can not find a TTI timing that is the exact median value of the runs, so it will return a ```undefined``` as the medianRun which generates the error seen in the screenshot below. By definition the median will average the middle 2 results.

If we merge https://github.com/paulirish/pwmetrics/pull/71 I also propose we change the way this function works to drop the TTI dependency which a user might not specify and which will crash the program.

![screen shot 2017-03-08 at 10 59 52](https://cloud.githubusercontent.com/assets/3718654/23701391/741cdda2-03ee-11e7-9f40-5edd9370ee0e.png)

paulirish:
to pick median run we need to pick SOME metric to sort on. :)

I started with TTI because it's the most composite of the metrics... but rather than that we could just find whatever had the longest value of all metrics all runs. and then use whatever that metric is to sort by.
I'm nearly certain it will always be TTI, normally. but that would make this subsetting more reasonable.

hows that sound?

pedro93:
Sounds good to me, with the desiredMetrics PR we may not always have tti metric so it would good to safeguard against it. 

Another question, what about when running pwmetrics with say 10 runs? 

There will not be a median value to pick from, as an average between the 5th and 6th (sorted) run will be choosen and that value has no match to the actual data retrieved.

paulirish:
> Another question, what about when running pwmetrics with say 10 runs?

afaik, this case is already handled.

pedro93:
Its not because of this line https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L205

Ignore the unexpected console.logs.

![screen shot 2017-03-10 at 17 42 45](https://cloud.githubusercontent.com/assets/3718654/23806286/05de9854-05b9-11e7-9293-fa90b7c02746.png)

pedro93:
@paulirish Any particular way you to want to handle finding the largest metric to sort by? I was thinking we could use the [metricsIds](https://github.com/paulirish/pwmetrics/blob/master/lib/metrics.js#L7) as a sorted (by biggest value) collection. 
We could also try some sort of dynamic calculation, not sure how to that atm though. 

[...]

Issue #55, opened 2017-02-08
Release 2.0


denar90:
- [x] TypeScript usage. Will help us to reduce number of tests and make lib more reliable. 
 - [x] Yargs - https://github.com/paulirish/pwmetrics/pull/51
 - [x] metricsDefinitions - https://github.com/paulirish/pwmetrics/pull/50 (after Lighthouse release)
 - [x] Handle CRI_TIMEOUT error - https://github.com/paulirish/pwmetrics/pull/42
 - [x] test sheets functionality - https://github.com/paulirish/pwmetrics/pull/38
 - [x] refactor/tests for [expectations](https://github.com/paulirish/pwmetrics/blob/master/lib/expectations.js) like it was done for sheets  https://github.com/paulirish/pwmetrics/pull/38
 - [x] tests for [recordLighthouseTrace](https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L79) postponed
 - [x] tests for [appendResults](https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L147) postponed
 - [x] tests for [findMedianRun](https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L175)
 - [x] refactor/tests for [gsheets](https://github.com/paulirish/pwmetrics/blob/master/lib/gsheets/gsheets.js) like it was done for sheets  https://github.com/paulirish/pwmetrics/pull/38
- [x] https://github.com/paulirish/pwmetrics/issues/53, https://github.com/paulirish/pwmetrics/issues/48 (postponed)
- [x] https://github.com/paulirish/pwmetrics/issues/28, https://github.com/paulirish/pwmetrics/issues/31 kinda critical, but depends on lighthouse fix https://github.com/GoogleChrome/lighthouse/pull/1672

Anything to add, thoughts

pedro93:
Yargs related tests.
Configuration file parameter validation.

There are also inconsistencies in the README about how to use the config file and certain flags uses. @denar90 if you would like I can talk to you over slack or something and explain.

denar90:
updated priority

pedro93:
We have promise chains without catch blocks which leads to hidden error messages. These should be cleaned up for a better error processing and logging.

E.g: If we catch an error [here](https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L133), we will not have access to a correct stack trace: 

[...]

PR #84, opened 2017-03-29
Tests find median run

pedro93:
As part of https://github.com/paulirish/pwmetrics/issues/55 addresses https://github.com/paulirish/pwmetrics/issues/72


lib/index.js
```diff
@@ -200,13 +199,29 @@ class PWMetrics {
...
+
+    values.sort(function sortValues(a, b) {
+      return a - b;
+    });
+
+    const half = Math.floor(values.length / 2);
+    if (values.length % 2)
+      return values[half];
+    else
+      return values[half]; // take slightly worse entry as median.
```
staabm: IF and ELSE branch contain the same return value?

pedro93: It's a mistake, thanks for catching it, just commited a fix. There was also the case where values.length is 2 and the half variable already is correct index.

lib/index.js
```diff
@@ -200,13 +199,29 @@ class PWMetrics {
...
     }));
     return medianRun;
   }
+
+  getMedianRun(values) {
+    if (values.length == 1) {
+      return values[0];
+    }
+
+    values.sort(function sortValues(a, b) {
```
denar90: just arrow func
`values.sort((a, b) => a-b);`

lib/index.js
```diff
@@ -200,13 +199,29 @@ class PWMetrics {
...
+      return a - b;
+    });
+
+    const half = Math.floor(values.length / 2);
+    if (values.length % 2 || values.length === 2)
+      return values[half];
+    else
+      return values[half+1]; // take slightly worse entry as median.
```
denar90: `half + 1`

lib/index.js
```diff
@@ -200,13 +199,29 @@ class PWMetrics {
...
+    const medianTTI = this.getMedianRun(ttiValues);
     // in the case of duplicate runs having the exact same TTI, we naively pick the first
     const medianRun = results.find(result => result.timings.find(timing => {
       return timing.id === metrics.ids.TTI && timing.timing === medianTTI;
     }));
     return medianRun;
   }
+
+  getMedianRun(values) {
+    if (values.length == 1) {
```
denar90: `===`

package.json
```diff
@@ -22,7 +22,6 @@
     "google-auth-library": "^0.10.0",
     "googleapis": "^16.0.0",
     "lighthouse": "1.6.0",
-    "median": "^0.0.2",
```
paulirish: we also have a yarn lock that'll need an update with this.

lib/index.js
```diff
@@ -200,13 +199,25 @@ class PWMetrics {
 
   findMedianRun(results) {
     const ttiValues = results.map(r => r.timings.find(timing => timing.id === metrics.ids.TTI).timing);
-    const medianTTI = median(ttiValues);
+    const medianTTI = this.getMedianRun(ttiValues);
     // in the case of duplicate runs having the exact same TTI, we naively pick the first
     const medianRun = results.find(result => result.timings.find(timing => {
```
paulirish: a little odd we now have `findMedianRun()` `getMedianRun()` and `medianRun`.

afaict, your new `getMedianRun` is just getting a median value of the TTIs passed in, but not returning a run.

can we clean this up?

i'm open to solutions, but one would be we nuke this last `results.find(result => result... ` bit and and pass the `results` into your new getMedianRun. In there, we operate on each `result` and compare its TTI result value to the other. and in the end we return the result object.

but if you want do it differently, that WFM too. :)

package.json
pedro93: Just committed a my proposal for the code change, when you have a minute let me know if you think it's ok @paulirish

lib/index.js
pedro93: Just committed my proposal for the code change, when you have a minute let me know if you think it's ok @paulirish

package.json
pedro93: done

lib/index.js
pedro93: done

lib/index.js
pedro93: done

lib/index.js
pedro93: done

lib/index.js
pedro93: done

lib/index.js
pedro93: done

PR #87, opened 2017-03-31
Use --view flag for timeline viewer

denar90:


denar90:
@paulirish sgtm, I wanted to look at readme in any case though :)

Issue #88, opened 2017-04-03
Process.exit


Bubbit:
What is the reason process.exit() is required at the end of the run? https://github.com/paulirish/pwmetrics/blob/master/recipes/gulp/gulpfile.js#L48
I'm trying to integrate pwmetrics as part of my CI-chain, but currently the only solution is to spawn it in a different process.

denar90:
We just wanted to and gulp task after `pwmetrics` finished it's job. But you can solve this case in other way, due to needs of your CI and gulp process.

Bubbit:
I'm trying to figure out which process isn't exiting properly as it should finish without me having to call a process.exit. I don't want to fix it by spawning a child process, but it's currently the only solution

denar90:
Does it exit correctly without `pwmetrics` ? Wanna figure out if problem caused by us.

Bubbit:
Yes it does exit correctly.

denar90:
Can you share code, plz?

Bubbit:
https://github.com/Bubbit/pwmetrics/blob/master/recipes/gulp/gulpfile.js

Doesn't close properly

denar90:
ok, I'll look at it but a bit later...

pedro93:
Hello @Bubbit thank you for posting this issue. 

I will have a look to try when I have more time and understand why, I am not sure if this is because of ```pwmetrics``` or because of ```lighthouse```. 
In the meantime I would suggest using ```child_process``` like so:

https://gist.github.com/pedro93/efb78fd310b1d1dcda656ad4aaabf413

Basically:
1.) Setup a local web server.
2.) Execute pwmetrics as a child process, give it a callback function so your server knows when the tests have concluded.
3.) Clean up the server.

pedro93:
@Bubbit @denar90 After digging around a little I found no particular reason why pwmetrics would maintain the process live. I checked in lighthouse and noticed the same happening: https://github.com/GoogleChrome/lighthouse/tree/master/docs/recipes/gulp

@Bubbit May I suggest you ask about ```process.exit(0)``` in the lighthouse repository? You can link back to this issue as needed. 

I will close this this comment with a will not fix as this is not of the responsability of pwmetrics as far as I can see. Feel free to reopen if you disagree.

Bubbit:
I filled the issue at the lighthouse repository, but it's not due to lighthouse. 

So I went back and did some more investigation, the process doesn't exit due to the chart output.
[...]

PR #89, opened 2017-04-04
Fix hanging process after chart draw

Bubbit:
Linked to https://github.com/paulirish/pwmetrics/issues/88

By destroying the chart after the draw the process closes and process.exit() is no longer needed


pedro93:
LGTM, will accept.

PR #92, opened 2017-04-06
index.js -> index.ts

denar90:
@paulirish @samccone last migration file to .ts 🍻 

P.S. Index file looks like newly created but if look at [blame history](https://github.com/denar90/pwmetrics/blame/ef26e6576786c52e7da0716a757b76410247b8bc/lib/index.ts), it's saved, I think it's a github diff issue...

Guys you can look commit after commit to watch diff better...

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -42,6 +42,7 @@ class PWMetrics {
     this.flags = Object.assign({}, this.flags, opts.flags);
     this.runs = this.flags.runs;
     this.sheets = opts.sheets;
+    this.expectations = opts.expectations;
     this.clientSecret = opts.clientSecret;
     this.tryLighthouseCounter = 0;
 
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -32,7 +32,7 @@ class PWMetrics {
     this.flags = opts.flags || {};
     // @todo remove when new lighthouse version be released, because -https://github.com/GoogleChrome/lighthouse/pull/1778
     this.flags.disableCpuThrottling = this.flags.disableCpuThrottling || false;
-    this.runs = opts.flags.runs || 1;
+    this.runs = this.flags.runs || 1;
     this.sheets = opts.sheets;
     this.expectations = opts.expectations || false;
     this.clientSecret = opts.clientSecret;
@@ -69,7 +69,7 @@ class PWMetrics {
       }
     }
 
-    results = results.filter(r => !(r instanceof Error));
+    const ret = {runs: results.filter(r => !(r instanceof Error))};
     if (this.runs > 1 && !this.flags.submit) {
       const median = this.findMedianRun(results);
       console.log(messages.getMessage('MEDIAN_RUN'));
@@ -78,7 +78,7 @@ class PWMetrics {
       const sheets = new Sheets(this.sheets, this.clientSecret);
       return sheets.appendResults(results).then(() => results);
     }
-    return results;
+    return ret;
   }
 
   async run(): Promise<MetricsResults> {
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -82,7 +82,7 @@ class PWMetrics {
 
     let results: PWMetricsResults = {runs: metricsResults.filter(r => !(r instanceof Error))};
     if (this.runs > 1 && !this.flags.submit) {
-      results.median = this.findMedianRun(metricsResults);
+      results.median = this.findMedianRun(results.runs);
       console.log(messages.getMessage('MEDIAN_RUN'));
       this.displayOutput(results.median);
     } else if (this.flags.submit) {
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -13,7 +13,7 @@ const expectations = require('./expectations');
 const {upload} = require('./upload');
 const messages = require('./utils/messages');
 
-import {MainOptions, FeatureFlags, AuthorizeCredentials, LighthouseResults, MetricsResults, DriveResponse, TermWritableStream, PWMetricsResults, SheetsConfig, ExpectationMetrics} from '../types/types';
+import {MainOptions, FeatureFlags, AuthorizeCredentials, LighthouseResults, MetricsResults, TermWritableStream, PWMetricsResults, SheetsConfig, ExpectationMetrics} from '../types/types';
 
 const MAX_LIGHTHOUSE_TRIES = 2;
 const SIGINT = 'SIGINT';
@@ -162,7 +162,7 @@ class PWMetrics {
       const preparedData = metrics.prepareData(data);
 
       if (this.flags.upload) {
-        const driveResponse: DriveResponse = await upload(data, this.clientSecret);
+        const driveResponse = await upload(data, this.clientSecret);
         this.view(driveResponse.id);
       }
 
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,10 +1,9 @@
 // Copyright 2016 Google Inc. All Rights Reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE
-'use strict';
 
 const lighthouse = require('lighthouse');
 const ChromeLauncher = require('lighthouse/lighthouse-cli/chrome-launcher');
-const perfConfig = require('lighthouse/lighthouse-core/config/perf.json');
+const perfConfig: any = require('lighthouse/lighthouse-core/config/perf.json');
 const opn = require('opn');
 
 const Sheets = require('./sheets/index');
@@ -12,20 +11,28 @@ const Chart = require('./chart/chart');
 const metrics = require('./metrics');
 const expectations = require('./expectations');
 const {upload} = require('./upload');
-const {getMessage, getMessageWithPrefix} = require('./utils/messages');
+const messages = require('./utils/messages');
+
+import {MainOptions, FeatureFlags, AuthorizeCredentials, LighthouseResults, MetricsResults, DriveResponse, TermWritableStream} from '../types/types';
 
 const _MAX_LIGHTHOUSE_TRIES = 2;
 const _SIGINT = 'SIGINT';
 const _SIGINT_EXIT_CODE = 130;
 
 class PWMetrics {
+  flags: FeatureFlags|any;
+  runs: number;
+  sheets: Object;
+  expectations: Object;
+  clientSecret: AuthorizeCredentials;
+  tryLighthouseCounter: number;
+  launcher: any;
 
-  constructor(url, opts) {
-    this.url = url;
+  constructor(public url: string, opts: MainOptions) {
     this.flags = opts.flags || {};
     // @todo remove when new lighthouse version be released, because -https://github.com/GoogleChrome/lighthouse/pull/1778
     this.flags.disableCpuThrottling = this.flags.disableCpuThrottling || false;
-    this.runs = this.flags.runs || 1;
+    this.runs = opts.flags.runs || 1;
     this.sheets = opts.sheets;
     this.expectations = opts.expectations || false;
     this.clientSecret = opts.clientSecret;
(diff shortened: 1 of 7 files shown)


lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  clientSecret: AuthorizeCredentials;
+  tryLighthouseCounter: number;
+  launcher: any;
+
+  constructor(public url: string, opts: MainOptions) {
+    this.flags = opts.flags || {};
+    // @todo remove when new lighthouse version be released, because -https://github.com/GoogleChrome/lighthouse/pull/1778
+    this.flags.disableCpuThrottling = this.flags.disableCpuThrottling || false;
+    this.runs = this.flags.runs || 1;
+    this.sheets = opts.sheets;
```
pedro93: No default value? Yargs does not have an default value for this either so it may be good for input validation to provide some fallback.

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+    this.expectations = opts.expectations || false;
+    this.clientSecret = opts.clientSecret;
```
pedro93: Same thing here?

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  async start() {
+    const runs = Array.apply(null, {length: this.runs}).map(Number.call, Number);
+    let results: MetricsResults[] = [];
+
+    // Kill spawned Chrome process in case of ctrl-C.
+    process.on(_SIGINT, async () => {
+      if (this.launcher) {
+        await this.launcher.kill();
```
pedro93: Won't this fail under some condition? Shouldn't it be wrapped in try catch?

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+        console.log(messages.getMessageWithPrefix('SUCCESS', 'SUCCESS_RUN', runIndex, runs.length));
+      } catch(error) {
+        results[runIndex] = error;
+        console.error(messages.getMessageWithPrefix('ERROR', 'FAILED_RUN', runIndex, runs.length, error.message));
+      }
+    }
+
+    const ret = {runs: results.filter(r => !(r instanceof Error))};
+    if (this.runs > 1 && !this.flags.submit) {
+      const median = this.findMedianRun(results);
```
pedro93: Change to 
```js
ret.median = this.findMedianRun(ret.runs)
....
this.displayOutput(ret.median);
```
If there is an error the findMedianRun function will not catch it and will crash the program.

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+      console.log(messages.getMessage('MEDIAN_RUN'));
+      this.displayOutput(median);
+    } else if (this.flags.submit) {
+      const sheets = new Sheets(this.sheets, this.clientSecret);
+      return sheets.appendResults(results).then(() => results);
```
pedro93: Same as above, use the ret variable:
```js
        return sheets.appendResults(ret.runs).then(_ => ret);
```

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+      try {
+        results[runIndex] = await this.run();
+        console.log(messages.getMessageWithPrefix('SUCCESS', 'SUCCESS_RUN', runIndex, runs.length));
+      } catch(error) {
+        results[runIndex] = error;
+        console.error(messages.getMessageWithPrefix('ERROR', 'FAILED_RUN', runIndex, runs.length, error.message));
+      }
+    }
+
+    const ret = {runs: results.filter(r => !(r instanceof Error))};
```
pedro93: Change to 
```js 
let
```
To permit the findMedianRun to add the median run.

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+
+      if (process.env.CI) {
+        // handling CRI_TIMEOUT issue - https://github.com/GoogleChrome/lighthouse/issues/833
+        this.tryLighthouseCounter = 0;
+        lhResults = await this.runLighthouseOnCI();
+      } else {
+        lhResults = await lighthouse(this.url, this.flags);
+      }
+
+      const metricsResults: MetricsResults = await this.recordLighthouseTrace(lhResults);
```
pedro93: Is this not inferable?

lib/index.ts
denar90: I we wanna validate all options came as argument then it should be standalone validator class...

lib/index.ts
denar90: I don't think we should do that in scope of this PR.

lib/index.ts
denar90: Yeap, I messed up with all this stuff. thx 👍

lib/index.ts
denar90: Here should be `await` stuff instead

lib/index.ts
denar90: what do you mean?

pedro93: recordLighthouseTrace returns a Promise with a value of type MetricsResults, can't that be used for any typescript inferrer to know what type that metricsResults is?
https://github.com/paulirish/pwmetrics/blob/master/lib/index.js#L134
https://github.com/paulirish/pwmetrics/blob/master/lib/metrics.ts#L37

denar90: `await this.recordLighthouseTrace(lhResults);` returns `MetricsResults`
`this.recordLighthouseTrace(lhResults)` returns `Promise<MetricsResults>`

pedro93: You are right, my bad 👍

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+const messages = require('./utils/messages');
+
+import {MainOptions, FeatureFlags, AuthorizeCredentials, LighthouseResults, MetricsResults, DriveResponse, TermWritableStream} from '../types/types';
+
+const _MAX_LIGHTHOUSE_TRIES = 2;
+const _SIGINT = 'SIGINT';
+const _SIGINT_EXIT_CODE = 130;
+
+class PWMetrics {
+  flags: FeatureFlags|any;
```
samccone: any?

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  runs: number;
+  sheets: Object;
+  expectations: Object;
+  clientSecret: AuthorizeCredentials;
+  tryLighthouseCounter: number;
+  launcher: any;
```
samccone: any :(

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+import {MainOptions, FeatureFlags, AuthorizeCredentials, LighthouseResults, MetricsResults, DriveResponse, TermWritableStream} from '../types/types';
+
+const _MAX_LIGHTHOUSE_TRIES = 2;
+const _SIGINT = 'SIGINT';
+const _SIGINT_EXIT_CODE = 130;
+
+class PWMetrics {
+  flags: FeatureFlags|any;
+  runs: number;
+  sheets: Object;
```
samccone: do we have no idea about the shape of these? I would prefer to avoid any + object

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+      } else {
+        lhResults = await lighthouse(this.url, this.flags);
+      }
+
+      const metricsResults: MetricsResults = await this.recordLighthouseTrace(lhResults);
+      await this.launcher.kill();
+
+      return metricsResults;
+    } catch (error) {
+      throw new Error(error);
```
samccone: do we need to rewrap the error?

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+      return this.launcher.kill();
+    }
+  }
+
+  async launchChrome() {
+    try {
+      this.launcher = new (ChromeLauncher.ChromeLauncher || ChromeLauncher)();
+      await this.launcher.isDebuggerReady();
+    } catch(e) {
+      console.log(messages.getMessage('LAUNCHING_CHROME'));
```
samccone: this log message seems out of place / different from what is going on?

lib/index.ts
```diff
@@ -0,0 +1,244 @@
...
+  median(values: Array<number>) {
+    if (values.length === 1) return values[0];
+    values.sort((a, b) => a - b);
+    const half = Math.floor(values.length / 2);
+    return values[half];
+  }
+
+  view(id: string) {
+    if (this.flags.view) {
+      opn(`https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=https://drive.google.com/file/d//${id}/view?usp=drivesdk`);
```
samccone: extract to file level const

lib/index.ts
denar90: I was trying to chit with ts build error for `this.flags = opts.flags || {};`

lib/index.ts
denar90: yeap, we don't, thx 👍

lib/index.ts
denar90: But it, kinda same to
```js
return this.launcher.isDebuggerReady()
      .catch(() => {
        console.log(getMessage('LAUNCHING_CHROME'));
        return this.launcher.run();
      });
```

isn't it?

samccone: but why is it in a catch? and not a .then ?

samccone: or are you saying...

if debugger is not ready, launch a new chrome so that we can have a debugger?

denar90: I suppose...

lib/index.ts
samccone: flags?: FeatureFlags

denar90: thx 👍

lib/index.ts
pedro93: Should we not kill the google chrome instance (this.launcher) if there is an error? If we don't, won't there be a zombie google chrome process? In the past I've had google chrome not closed because of a failed run.

lib/index.ts
```diff
@@ -13,28 +13,35 @@ const expectations = require('./expectations');
...
-const _SIGINT = 'SIGINT';
-const _SIGINT_EXIT_CODE = 130;
+const MAX_LIGHTHOUSE_TRIES = 2;
+const SIGINT = 'SIGINT';
+const SIGINT_EXIT_CODE = 130;
+const getTimelineViewerUrl = (id: string) => `https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=https://drive.google.com/file/d//${id}/view?usp=drivesdk`
 
 class PWMetrics {
-  flags: FeatureFlags|any;
+  flags: FeatureFlags = {
```
denar90: @samccone I've decided to setup default values as final solution.

lib/index.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    // Kill spawned Chrome process in case of ctrl-C.
+    process.on(SIGINT, async() => {
+      if (this.launcher) {
+        try {
+          await this.killLauncher();
+          process.exit(SIGINT_EXIT_CODE);
+          console.log(messages.getMessage('CLOSING_CHROME'));
+        } catch (error) {
+          throw error;
```
pedro93: Should we throw an error? If yes then why do we catch the error and rethrow it?

lib/index.ts
```diff
@@ -0,0 +1,261 @@
...
+        console.log(messages.getMessageWithPrefix('SUCCESS', 'SUCCESS_RUN', runIndex, runs.length));
+      } catch (error) {
+        metricsResults[runIndex] = error;
+        console.error(messages.getMessageWithPrefix('ERROR', 'FAILED_RUN', runIndex, runs.length, error.message));
+      }
+    }
+
+    let results: PWMetricsResults = {runs: metricsResults.filter(r => !(r instanceof Error))};
+    if (this.runs > 1 && !this.flags.submit) {
+      results.median = this.findMedianRun(metricsResults);
```
pedro93: ```metricsResults``` are not error filtered, change to ```results.runs```

lib/index.ts
```diff
@@ -0,0 +1,262 @@
...
+
+    if (this.flags.expectations) {
+      if (this.expectations) {
+        expectations.validateMetrics(this.expectations);
+        this.expectations = expectations.normalizeMetrics(this.expectations);
+      } else throw new Error(getMessageWithPrefix('ERROR', 'NO_EXPECTATIONS_FOUND'));
+    }
+  }
+
+  async start() {
```
pedro93: should start() explicitly return a ```Promise<PWMetricsResults>```?

lib/index.ts
```diff
@@ -0,0 +1,262 @@
...
+    const medianTTI = this.median(ttiValues);
+    // in the case of duplicate runs having the exact same TTI, we naively pick the first
+    // @fixme, but any for now...
+    return results.find((result: any) => result.timings.find((timing:any) =>
+        timing.id === metrics.ids.TTI && timing.timing === medianTTI
+      )
+    );
+  }
+
+  median(values: Array<number>) {
```
pedro93:
```js
median(values: Array<number>): number
```
Or is Typescript smart enough to detect that?

lib/index.ts
denar90: >If yes then why do we catch the error and rethrow it?

what do you mean? do you wanna rethrow it in other place?

lib/index.ts
denar90: Yeap, thx 👍

lib/index.ts
denar90: From docs
>We can add types to each of the parameters and then to the function itself to add a return type. TypeScript can figure the return type out by looking at the return statements, so we can also optionally leave this off in many cases.

I think it's smart though...

lib/index.ts
pedro93: No, I mean remove the try catch. Not sure if an await throws an error inside a promise function is the equivalent to Promise.reject(error).

Would this work?
```js
process.on(SIGINT, async() => {
      if (this.launcher) {
          await this.killLauncher();
          process.exit(SIGINT_EXIT_CODE);
          console.log(messages.getMessage('CLOSING_CHROME'));
      }
    });
```

Same thing for https://github.com/denar90/pwmetrics/blob/e98b78aa2039c174f381400778c77ea49263bcaf/lib/index.ts#L114 and https://github.com/denar90/pwmetrics/blob/e98b78aa2039c174f381400778c77ea49263bcaf/lib/index.ts#L180

denar90: Way to catch errors for async/await is using try/catch
https://tc39.github.io/ecmascript-asyncawait/#examples

lib/index.ts
samccone: typescript can infer this, no need to set it manually

denar90:
@pedro93 can we land this one?

pedro93:
Go right ahead 👍

denar90:
Landed 💥 
@samccone @pedro93 thx for good review, not really good stuff was from me. 
🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻 🍻  then!

PR #96, opened 2017-04-13
Update to 1.2.5 to make internal npm registries happy

Bubbit:
There have been some updates recently, but the version has remained the same, internally at our company we use Artifactory, which means these changes won't be passed down as it's cached :(


denar90:
@Bubbit we gonna release v2 in near future, but if you wanna use it now you should install it from master `npm i paulirish/pwmetrics --save-dev`
I think release 1.2.5 will not happen...

Bubbit:
Ok, for now we'll have to use the master branch, only risk is that we use in a CI pipeline so version are easier to manage than a branch that can update at any time

denar90:
You can stick to special commit
if I'm not mistaken - `npm i paulirish/pwmetrics#commit --save-dev`

PR #98, opened 2017-04-13
Fix CI usage

### LINKED ISSUE #63: CI issue after updating lighthouse version
We have issue on CI after finishing measurements.
https://travis-ci.org/paulirish/pwmetrics/jobs/204950872#L722-L735

So I found that it's related to killing launched chrome.
Everything is fine if we don't kill it.

Here is my commit with changes.
I've removed all logic related to charts and stuff.

Diff - https://github.com/paulirish/pwmetrics/compare/master...denar90:dont-kill-chrome
CI - https://travis-ci.org/denar90/pwmetrics/jobs/205395193

Then I downgraded version to `1.4.1` and it works perfectly

Diff - https://github.com/paulirish/pwmetrics/compare/master...denar90:test-ci-with-1.4.1
CI - https://travis-ci.org/denar90/pwmetrics/jobs/205392241

I suppose that something has changed in chrome launcher in `1.5.x`...

cc @paulirish @samccone

denar90:
Resolves #63

### DIFF
--- a/recipes/gulp/gulpfile.js
+++ b/recipes/gulp/gulpfile.js
@@ -23,7 +23,6 @@ const startServer = function() {
  */
 const stopServer = function() {
   connect.serverClose();
-  process.exit(0);
 };
 
 /**
--- a/recipes/gulp/gulpfile.js
+++ b/recipes/gulp/gulpfile.js
@@ -23,6 +23,7 @@ const startServer = function() {
  */
 const stopServer = function() {
   connect.serverClose();
+  process.exit(0);
 };
 
 /**
@@ -57,7 +58,8 @@ const runPwmetrics = function() {
  */
 const handleOk = function(results) {
   stopServer();
-  return results;
+  console.log(results);
+  process.exit(0);
 };
 
 /**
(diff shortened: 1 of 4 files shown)


recipes/gulp/gulpfile.js
```diff
@@ -66,7 +67,7 @@ const handleOk = function(results) {
 const handleError = function(e) {
   stopServer();
   console.error(e); // eslint-disable-line no-console
-  throw e; // Throw to exit process with status 1.
+  process.exit(1);
```
pedro93: Why do we need this process.exit, is the ```throw``` not enough?

recipes/gulp/gulpfile.js
```diff
@@ -57,7 +57,8 @@ const runPwmetrics = function() {
  */
 const handleOk = function(results) {
   stopServer();
-  return results;
+  console.log(results);
+  process.exit(0);
```
pedro93: Is this necessary?

denar90: yeap, because travis will not be closed. I noticed that during build https://travis-ci.org/denar90/pwmetrics/jobs/221887465
And we wanna pass our recipe which is also like integration test

recipes/gulp/gulpfile.js
denar90: Just to be consistent with finishing run in `handleOk` and `handleError` methods

pedro93:
LGTM

PR #102, opened 2017-04-19
Use push(true) and pop(true) everywhere to fix Mac Terminal positioning

exogen:
This includes the flag to enable `withAttributes` on all `push()` and `pop()` calls to `charm`. For whatever reason, this fixes alignment in macOS Terminal.

screen shot 2017-04-19 at 1 30 06 pm

(This would be possible to install/test directly from this branch if [postinstall-build](https://github.com/exogen/postinstall-build) were used 😉)

### DIFF
--- a/lib/chart/bar.js
+++ b/lib/chart/bar.js
@@ -16,10 +16,10 @@ Bar.prototype.draw = function(scale) {
   const charm = this.chart.charm;
   const dir = this.chart.direction;
   if (dir === 'x' && this.label) {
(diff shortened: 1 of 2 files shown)


denar90:
@exogen thanks, I've rechecked for mac terminal and iterm, both looks fine 
Alt text
Alt text

@paulirish I think it can feat v2.

paulirish:
awesome. thank you so much for figuring this out! :)

PR #110, opened 2017-04-25
Bump ts to 2.3, fixes

denar90:


lib/metrics.ts
```diff
@@ -26,6 +28,8 @@ module.exports = {
     metricsIds.VC85,
     metricsIds.TTInonvis,
     metricsIds.TTInonvis5s,
+    metricsIds.EndOfATrace,
```
denar90: I found out that for some sites `eot` and `onload ` are shown on a chart...

samccone:
👍 lgtm

paulirish:
👍  on shipping 2.0.1 btw with this.

paulirish:
Ah, you already did it! Ace.

PR #112, opened 2017-05-04
Config fix

### LINKED ISSUE #111: Cannot find module '/my/path/to/package.json'
After running `pwmetrics` from path where `package.json` is absent we get exception kinda `Cannot find module '/my/path/to/package.json'`
Thanks to @pedro93 for finding this one 👍

denar90:
Resolves #111

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,6 +9,7 @@
 ![image](https://cloud.githubusercontent.com/assets/39191/19417867/7aead922-93af-11e6-88ec-917dad6e89d2.png)
 
  Documentation on these metrics in the works. If you hit bugs in the metrics collection, report at [Lighthouse issues](https://github.com/GoogleChrome/lighthouse/issues).
+ [How to use article](https://medium.com/@denar90/easy-progressive-web-metrics-9afa5ed857c2)
 
 ### Install [![NPM pwmetrics package](https://img.shields.io/npm/v/pwmetrics.svg)](https://npmjs.org/package/pwmetrics)
 ```sh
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -9,6 +9,7 @@
 ![image](https://cloud.githubusercontent.com/assets/39191/19417867/7aead922-93af-11e6-88ec-917dad6e89d2.png)
 
  Documentation on these metrics in the works. If you hit bugs in the metrics collection, report at [Lighthouse issues](https://github.com/GoogleChrome/lighthouse/issues).
+ [How to use article](https://medium.com/@denar90/easy-progressive-web-metrics-9afa5ed857c2)
```
denar90: Don't you mind guys if I add the article about `pwmetrics` usage?

pedro93:
@denar90 can you create tests at the cmd line tool level (https://github.com/paulirish/pwmetrics/blob/master/test/cli.js) to ensure this case does not break in the future? I agree with @paulirish there should be a cleaner/ less convoluted way to do this? I've haven't had the time to think about this but I'll try during the weekend. If I can't find an alternative and we get more tests for this then LGTM

denar90:
I agree with you guys that it looks like overhead. The reason I did it to simplify reading the code. If `package.json` => go one way, read just function, if file => read another function. Don't need to follow some conditions and stuff. I'm also open for improving it :)

pedro93:
@paulirish @denar90 I modified the PR code with my proposed changes, the relevant part is in the bin/cli.js file. What do you guys think?

paulirish:
LGTM.  yeah this is great.

over to @denar90 for confirmation this solves #111

denar90:
Nice solution 👍

test/cli.js
```diff
@@ -2,24 +2,40 @@
...
       try {
-        childProcess.execSync('node bin/cli.js --config=./test/fixtures/empty-config.js');
+        childProcess.execSync('bin/cli.js --config=./test/fixtures/empty-config.js');
       } catch (e) {
         expect(e.message).to.contain('No url entered..');
       }
     });
+
+    it('should not throw any error if a url is provided from cli', () => {
+      return childProcess.exec('bin/cli.js https://example.com', error => {
```
denar90: Sam is working on https://github.com/GoogleChrome/lighthouse/issues/2092, so it should fix our problem in future. But for now, we can make some ugly fix.

PR #120, opened 2017-05-16
LHv2

denar90:
I've done some experiments with hiding TTI, showing TTFI and TTCI. Also integrated stuff with updated chrome launcher

Trace of nytimes
image

Stil can't make yarn and my fork works, have to manually build stuff in chrome launcher folder.

Also, we have to update the link to new sheets in doc example because of new metrics.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -156,10 +156,12 @@ class PWMetrics {
   async launchChrome(): Promise<LaunchedChrome|Error> {
     try {
       console.log(messages.getMessage('LAUNCHING_CHROME'));
-      return this.launcher = await launch({
+      this.launcher = await launch({
         port: this.flags.port,
         chromeFlags: this.flags.chromeFlags
       });
+      this.flags.port = this.launcher.port;
+      return this.launcher;
     } catch(error) {
       await this.killLauncher();
       return error;
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See LICENSE
 
 const lighthouse = require('lighthouse');
-import {ChromeLauncher} from 'lighthouse/lighthouse-cli/chrome-launcher';
+import {launch, LaunchedChrome} from 'lighthouse/chrome-launcher/chrome-launcher';
 const perfConfig: any = require('lighthouse/lighthouse-core/config/perf.json');
 const opn = require('opn');
 
@@ -16,8 +16,6 @@ const messages = require('./utils/messages');
 import {MainOptions, FeatureFlags, AuthorizeCredentials, LighthouseResults, MetricsResults, TermWritableStream, PWMetricsResults, SheetsConfig, ExpectationMetrics} from '../types/types';
 
 const MAX_LIGHTHOUSE_TRIES = 2;
-const SIGINT = 'SIGINT';
-const SIGINT_EXIT_CODE = 130;
 const getTimelineViewerUrl = (id: string) => `https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=https://drive.google.com/file/d//${id}/view?usp=drivesdk`
 
 class PWMetrics {
@@ -29,14 +27,15 @@ class PWMetrics {
     expectations: false,
     output: false,
     // @todo remove when new lighthouse version be released, because -https://github.com/GoogleChrome/lighthouse/pull/1778
-    disableCpuThrottling: false
+    disableCpuThrottling: false,
+    chromeFlags: []
   };
   runs: number;
   sheets: SheetsConfig;
   expectations: ExpectationMetrics;
   clientSecret: AuthorizeCredentials;
   tryLighthouseCounter: number;
-  launcher: ChromeLauncher | undefined = undefined;
+  launcher: LaunchedChrome | undefined;
 
   constructor(public url: string, opts: MainOptions) {
     this.flags = Object.assign({}, this.flags, opts.flags);
@@ -58,18 +57,6 @@ class PWMetrics {
     const runs = Array.apply(null, {length: +this.runs}).map(Number.call, Number);
     let metricsResults: MetricsResults[] = [];
 
-    // Kill spawned Chrome process in case of ctrl-C.
-    process.on(SIGINT, async() => {
-      if (this.launcher) {
-        try {
-          await this.killLauncher();
-          process.exit(SIGINT_EXIT_CODE);
-          console.log(messages.getMessage('CLOSING_CHROME'));
-        } catch (error) {
-          throw error;
-        }
-      }
-    });
     for (let runIndex of runs) {
       try {
         metricsResults[runIndex] = await this.run();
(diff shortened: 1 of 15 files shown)


lib/index.ts
```diff
@@ -158,15 +143,17 @@ class PWMetrics {
...
-      await this.launcher.run();
-      return this.launcher;
+      this.launcher = await launch({
+        port: this.flags.port,
+        chromeFlags: this.flags.chromeFlags,
+        handleSIGINT: true
+      });
+      this.flags.port = this.launcher.port;
+    } catch(error) {
+      return error;
```
paulirish: probably want to do `this.launcher.kill()` in the catch

lib/index.ts
```diff
@@ -158,15 +143,17 @@ class PWMetrics {
...
-      await this.launcher.isDebuggerReady();
-      return this.launcher;
-    } catch(e) {
       console.log(messages.getMessage('LAUNCHING_CHROME'));
-      await this.launcher.run();
-      return this.launcher;
+      this.launcher = await launch({
+        port: this.flags.port,
+        chromeFlags: this.flags.chromeFlags,
+        handleSIGINT: true
```
paulirish: this one is the default so you can omit it.

package.json
```diff
@@ -21,7 +21,7 @@
     "charm": "^1.0.2",
     "google-auth-library": "^0.10.0",
     "googleapis": "^16.0.0",
-    "lighthouse": "paulirish/lighthouse#c74ea959",
+    "lighthouse": "denar90/lighthouse#14cd4f5",
```
paulirish: you can do  

    "lighthouse": "canary",

or 

    "lighthouse": "2.0.0-alpha.6",

lib/index.ts
samccone: +1

lib/index.ts
```diff
@@ -28,15 +26,14 @@ class PWMetrics {
     view: false,
     expectations: false,
     output: false,
-    // @todo remove when new lighthouse version be released, because -https://github.com/GoogleChrome/lighthouse/pull/1778
```
pedro93: @paulirish Will there be no way to disable cpu throttling in lighthouse v2?

denar90: The flag still present but
In LH v1 `disableCpuThrottling` is set up as `true` by default, we wanted it to be false.
In LH v2 `disableCpuThrottling` is `false` by default.

denar90:
Only thing is left is error after running `yarn`
```
> pwmetrics@2.0.2 build /home/travis/build/paulirish/pwmetrics
> tsc

node_modules/lighthouse/chrome-launcher/utils.ts(22,25): error TS7016: Could not find a declaration file for module 'mkdirp'. '/home/travis/build/paulirish/pwmetrics/node_modules/mkdirp/index.js' implicitly has an 'any' type.

```

I still have to do `cd node_modules/lighthouse/chrome-launcher && yarn` then it works...

pedro93:
Can this not be fixed by a postinstall step on lighthouse? I think it makes sense to have for the chrome-launcher and lighthouse-core modules (assuming we need any typescript files)

lib/index.ts
paulirish: yes what @denar90 said.   we now have cpu throttling **on** as a default
and yeah you can easily disable that.

paulirish:
> I still have to do cd node_modules/lighthouse/chrome-launcher && yarn then it works...

yes. good point.  noted that here: https://github.com/GoogleChrome/lighthouse/issues/2251#issuecomment-302128158

agree that it's kinda broken for typescript consuming apps right now.
cc @samccone

package.json
julianusti: 👍

lib/index.ts
```diff
@@ -158,15 +153,16 @@ class PWMetrics {
...
     try {
-      this.launcher = new ChromeLauncher();
-      await this.launcher.isDebuggerReady();
-      return this.launcher;
-    } catch(e) {
       console.log(messages.getMessage('LAUNCHING_CHROME'));
-      await this.launcher.run();
-      return this.launcher;
+      return this.launcher = await launch({
+        port: this.flags.port,
```
pedro93: this.flags.port is never set is it?

denar90: do we have to set default port? I thought chrome launcher has to take care of it...

pedro93: I'm not sure, that particular line was there from the original PR code, there was a cyclic dependency on ports. I'll take a look soon

samccone: no need to set a default port :)

pedro93: Tried it locally but I get this:
```
Error: connect ECONNREFUSED 127.0.0.1:9222
    at Object.exports._errnoException (util.js:1026:11)
    at exports._exceptionWithHostPort (util.js:1049:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1136:14)
```

Chrome opens a window with a tab but nothing happens.

denar90: +1 faced the same issue. Using some other port - works perfectly.
`pwmetrics http://example.com --port=8080`

samccone: If you use a dynamic port (no port flag) you need to use the port value that is passed back from launcher to connect with. 

I would be you or lighthouse has 9222 hard coded somewhere that you are not overriding.

pedro93: Yes we missed that, I added the port to lighthouse. I think it's best let chrome-launcher find an available port and use that to prevent port-in-use errors. @samccone Lighthouse is port-agnostic too right? It should work on any available port correct?

denar90: @samccone thx, btw wdyt about @pedro93 idea?

paulirish: @pedro93 yup chrome-launcher automatically finds an available port and uses that. LH is port agnostic yup.

what's on the branch now is how i'd do it!

denar90:
@paulirish shouldn't we clean up https://github.com/paulirish/pwmetrics/blob/master/lib/metrics.ts#L25 due to 8519e27 ?

pedro93:
Because lighthouse v2 no longer has the TTI metric we need to pick a new to calculate the median run. See https://github.com/denar90/pwmetrics/blob/8519e2713987d5ed24af906e3a28fc71e0d421ef/lib/index.ts#L248 and https://github.com/denar90/pwmetrics/blob/8519e2713987d5ed24af906e3a28fc71e0d421ef/lib/index.ts#L253 I've just run it this problem. @paulirish What is the closest metric to TTI now? TTFI? Should we use an older more stable metric instead?

pedro93:
Running pwmetrics on https://github.com gives out the following error:
```
  ✘ Error: Unable to complete run 1 of 1 due to Audit error: The main thread was busy for the entire trace recording. First Interactive requires the main thread to be idle for several seconds. Audit key: first-interactive.
```

Using lighthouse version 2.1.0 installed globally via npm, with node v.8.0.0 this metric is correctly obtained. I don't think we should merge until we find out why we get this error.

denar90:
Guys, I think we are ready to ship this one 🚢

pedro93:
Be my guest @denar90 :)

pedro93:
Should this be pwmetrics v3?

paulirish:
Probably yeah

denar90:
Sorry, no squash.

denar90:
Aaaaand published 💥

PR #125, opened 2017-06-23
reduce eslint warnings

paulirish:
from 126 warnings to 18 warnings. \o/


denar90:
wow cool! what small changes reduce so many warnings \0/

Issue #127, opened 2017-07-10
pwmetrics returns `exit 0` when one of the expectations is not correct


willmendesneto:
### About this issue

I'm using pwmetrics, but I'm receiving a false positive when one of the expectations are passing. I couldn't find a way to make it pass only when all expectations are correct in the docs.

### Config / CLI options

```javascript
module.exports = {
  url: 'http://localhost:3000',
  flags: {
    runs: '1',
    submit: false,
    upload: false,
    view: false,
    expectations: true,
    disableCpuThrottling: false,
    chromeFlags: [
      '--incognito',
      '--aggressive-cache-discard',
      '--headless',
    ],
  },
  expectations: {
    ttfcp: {
      warn: '>=1500',
      error: '>=3000',
    },
    ttfmp: {
      warn: '>=2000',
      error: '>=4000',
    },
    fv: {
      warn: '>=1500',
      error: '>=4000',
    },
    psi: {
      warn: '>=1500',
      error: '>=3000',
    },
    vc85: {
      warn: '>=1500',
      error: '>=3000',
    },
    vs100: {
      warn: '>=1500',
      error: '>=4000',
    },
    tti: {
      warn: '>=1500',
      error: '>=4000',
    },
  },
};
```

### Steps to reproduce.

I tried to make it run using 

```shell
# run-pwmetrics.sh
#!/usr/bin/env bash
set -e
set -o pipefail

pwmetrics http://localhost:3000/landing --config=./pwmetrics-config.js 2> /dev/null
OUT=$?

echo "OUT: ${OUT}"
if [ $OUT -eq 0 ]
then
  echo "Successfully"
else
  echo "Error" >&2
fi
```

```bash
$ ./run-pwmetrics.sh
```

PS:

- I tried run via npm script adding the same command as a npm script and both return the same error.

### Screenshot

![screen shot 2017-07-10 at 3 09 49 pm](https://user-images.githubusercontent.com/1252570/28003554-e13a4cfe-6581-11e7-91fa-76d0738dfa0a.png)

### Environment

1. `pwmetrics` version: 2.0.2
2. Chrome version: Version 59.0.3071.115 (Official Build) (64-bit)
3. OS version: macOS Sierra 10.12.5 
4. NodeJS version: v6.10.0
5. Yarn version: v0.27.5

pedro93:
As of right now the status code of pwmetrics is relative to a successful auditing execution and not the values of that execution. 

[...]

PR #128, opened 2017-07-10
throw error when expectation is failing

willmendesneto:
### About this pull request

pwmetrics validation is not failing since it's relative to a successful auditing execution and not the values of that execution. These changes are adding this feature so that it can be used for validations and metrics more assertive.

### Points for share

- Solved: `NO_EXPECTATIONS_FOUND` was not using `messages`;
- Added new message `HAS_EXPECTATION_ERRORS` to be returned in case of result has expectations with errors using the flag `--expectations` on CLI or `expectations: true` via configuration;
-  Adding tests for validating new feature;

Pwmetrics issue: https://github.com/paulirish/pwmetrics/issues/127

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -88,6 +88,7 @@ class PWMetrics {
 
       if (this.resultHasExpectationErrors(results) && this.flags.expectations) {
         throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
+        process.exit(1);
       }
     }
     return results;
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -57,7 +57,7 @@ class PWMetrics {
       if (this.expectations) {
         expectations.validateMetrics(this.expectations);
         this.expectations = expectations.normalizeMetrics(this.expectations);
-      } else throw new Error(getMessageWithPrefix('ERROR', 'NO_EXPECTATIONS_FOUND'));
+      } else throw new Error(messages.getMessageWithPrefix('ERROR', 'NO_EXPECTATIONS_FOUND'));
     }
   }
 
@@ -85,8 +85,36 @@ class PWMetrics {
         const sheets = new Sheets(this.sheets, this.clientSecret);
         await sheets.appendResults(results.runs);
       }
+
+      if (this.resultHasExpectationErrors(results) && this.flags.expectations) {
+        throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
+      }
     }
     return results;
+  }
+
+  resultHasExpectationErrors(results: PWMetricsResults): boolean {
+    const expectationErrors = results.runs.filter(run => {
+      return run.timings.filter(item => {
+        const metricName = item.id;
+        const isAMetricErrorTiming = this.isAMetricErrorTiming(
+          item.timing,
+          metricName
+        );
+        return isAMetricErrorTiming === true;
+      });
+    });
+
+    return expectationErrors.length > 0;
+  }
+
+  isAMetricErrorTiming(metricValue: any, metricName: string): boolean {
+    const expectation = this.expectations[metricName];
+    if (expectation) {
+      const expectedErrorLimit = expectation.error;
+      return expectedErrorLimit !== undefined && metricValue >= expectedErrorLimit;
+    }
+    return false;
   }
 
   async run(): Promise<MetricsResults> {
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -22,7 +22,8 @@ import {
   TermWritableStream,
   PWMetricsResults,
   SheetsConfig,
-  ExpectationMetrics
+  ExpectationMetrics,
+  Timing
 } from '../types/types';
 
 const MAX_LIGHTHOUSE_TRIES = 2;
@@ -67,12 +68,20 @@ class PWMetrics {
 
     for (let runIndex of runs) {
       try {
-        metricsResults[runIndex] = await this.run();
+        const currentMetricResult: MetricsResults = await this.run();
+        currentMetricResult.error = this.resultHasExpectationErrors(currentMetricResult);
+        metricsResults[runIndex] = currentMetricResult;
         console.log(messages.getMessageWithPrefix('SUCCESS', 'SUCCESS_RUN', runIndex, runs.length));
       } catch (error) {
         metricsResults[runIndex] = error;
         console.error(messages.getMessageWithPrefix('ERROR', 'FAILED_RUN', runIndex, runs.length, error.message));
       }
+    }
+
+    const hasMetricResultsError = metricsResults.filter(m => !m.error).length > 0;
+
+    if (hasMetricResultsError && this.flags.expectations) {
+      throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
     }
 
     let results: PWMetricsResults = { runs: metricsResults.filter(r => !(r instanceof Error)) };
@@ -85,26 +94,14 @@ class PWMetrics {
         const sheets = new Sheets(this.sheets, this.clientSecret);
         await sheets.appendResults(results.runs);
       }
-
-      if (this.resultHasExpectationErrors(results) && this.flags.expectations) {
-        throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
-      }
     }
     return results;
   }
 
-  resultHasExpectationErrors(results: PWMetricsResults): boolean {
-    const expectationErrors = results.runs.filter(run => {
-      return run.timings.filter(item => {
-        const metricName = item.id;
-        const isAMetricErrorTiming = this.isAMetricErrorTiming(
-          item.timing,
-          metricName
-        );
-        return isAMetricErrorTiming === true;
-      });
+  resultHasExpectationErrors(metrics: MetricsResults): boolean {
+    const expectationErrors = metrics.timings.filter(timing => {
+      return !!this.isAMetricErrorTiming(timing.timing, timing.id);
     });
-
     return expectationErrors.length > 0;
   }
 
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -101,19 +101,15 @@ class PWMetrics {
   }
 
   resultHasExpectationErrors(metrics: MetricsResults): boolean {
-    const expectationErrors = metrics.timings.filter(timing => {
-      return !!this.isAMetricErrorTiming(timing.timing, timing.id);
+    const expectationErrors = metrics.timings.filter((timing: Timing) => {
+      const expectation = this.expectations[timing.id];
+      if (!expectation) {
+        return false;
+      }
+      const expectedErrorLimit = expectation.error;
+      return expectedErrorLimit !== undefined && timing.timing >= expectedErrorLimit;
     });
     return expectationErrors.length > 0;
-  }
-
-  isAMetricErrorTiming(metricValue: any, metricName: string): boolean {
-    const expectation = this.expectations[metricName];
-    if (expectation) {
-      const expectedErrorLimit = expectation.error;
-      return expectedErrorLimit !== undefined && metricValue >= expectedErrorLimit;
-    }
-    return false;
   }
 
   async run(): Promise<MetricsResults> {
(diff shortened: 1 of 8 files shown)


lib/index.ts
```diff
@@ -57,7 +57,7 @@ class PWMetrics {
       if (this.expectations) {
         expectations.validateMetrics(this.expectations);
         this.expectations = expectations.normalizeMetrics(this.expectations);
-      } else throw new Error(getMessageWithPrefix('ERROR', 'NO_EXPECTATIONS_FOUND'));
+      } else throw new Error(messages.getMessageWithPrefix('ERROR', 'NO_EXPECTATIONS_FOUND'));
```
paulirish: nice find! :)

lib/index.ts
```diff
@@ -85,10 +85,38 @@ class PWMetrics {
         const sheets = new Sheets(this.sheets, this.clientSecret);
         await sheets.appendResults(results.runs);
       }
+
+      if (this.resultHasExpectationErrors(results) && this.flags.expectations) {
+        throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
```
paulirish: we may have consumers of pwmetrics the node module, but we also have CLI users.  We want this to make sense for both.

On the CLI we don't want a big JS stack, so we want to catch the exception in the CLI, output a relevant error message, and `process.exit(1);`. For the node side, this seems ok.

lib/index.ts
```diff
@@ -85,10 +85,38 @@ class PWMetrics {
...
+      }
     }
     return results;
   }
 
+  resultHasExpectationErrors(results: PWMetricsResults): boolean {
+    const expectationErrors = results.runs.filter(run => {
+      return run.timings.filter(item => {
```
paulirish: this is a pretty heavyweight way to dive back into the results and expectations. Can we instead set a flag on the state of the run while `checkExpectations` runs?

willmendesneto: I totally agree that's pretty heavy do that, but `checkExpectations` uses `Timing` interface and add some flag there creates a cascade step if I understood correctl (PWMetricsResults >> MetricsResults >> Timing). Could you elaborate a little bit more about it, please?

pedro93: What if instead of looping through every result for a missed expectation you do it immediately after each run?
https://github.com/willmendesneto/pwmetrics/blob/c41b753f6dbdcac8da15916151a0907134cfabd0/lib/index.ts#L70

Something like
```js
try {
 const lhMetric: MetricsResults = await this.run();
 lhMetric.timings
   .filter(metric => metric.timing >= expectations[metric.id])
   .forEach(metric => {
     // Here you should have access to each failed metric for a specific run.
     // You can set a flag to false as the end state (if already false dont touch) 
     // like @paulirish suggested, there is probably a more performant method than looping
     // through every timing in every run.
   });
  // check if any failed, filter(...).length > 0 ? and return accordingly?
 metricsResults[runIndex] = lhMetric;
 console.log(messages.getMessageWithPrefix('SUCCESS', 'SUCCESS_RUN', runIndex, runs.length));
}
```

I do have a question, if you ask pwmetrics to execute n runs, should you fail on the first not meet expectation, should you test the expectations only for the median run or something in between?

lib/index.ts
```diff
@@ -85,10 +85,38 @@ class PWMetrics {
...
       }
+
+      if (this.resultHasExpectationErrors(results) && this.flags.expectations) {
+        throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
+      }
     }
     return results;
   }
 
+  resultHasExpectationErrors(results: PWMetricsResults): boolean {
```
denar90: what if we move this method in scope of https://github.com/paulirish/pwmetrics/blob/master/lib/expectations.ts ?

willmendesneto: I got your concern, but it's not valid in that case because we are working with `PWMetricsResults` which is related to `PWMetrics` file.

lib/index.ts
willmendesneto: Thanks for the explanation @pedro93 . Regarding your question, it should fail after execute all run commands.

willmendesneto:
@paulirish @pedro93 @denar90 Could you guys take a look again, please? Thanks!

lib/index.ts
```diff
@@ -67,14 +68,22 @@ class PWMetrics {
 
     for (let runIndex of runs) {
       try {
-        metricsResults[runIndex] = await this.run();
+        const currentMetricResult: MetricsResults = await this.run();
+        currentMetricResult.error = this.resultHasExpectationErrors(currentMetricResult);
```
pedro93: Why not simply declare a boolean outside of the loop? This remove the need to filter the result array in line 81. You could go a step further and check if the boolean has already been set by resultHasExpectationErrors then don't call it again and skip processing.

pedro93: I see you created 2 methods, resultHasExpectationErrors and isAMetricErrorTiming. The first one is really just a 1 line loop through the MetricsResults, I don't think it makes sense to have a function that simply looks an array to call a function. I would suggest you merge both new functions into one.

lib/index.ts
```diff
@@ -67,14 +68,22 @@ class PWMetrics {
...
       } catch (error) {
         metricsResults[runIndex] = error;
         console.error(messages.getMessageWithPrefix('ERROR', 'FAILED_RUN', runIndex, runs.length, error.message));
       }
     }
 
+    const hasMetricResultsError = metricsResults.filter(m => !m.error).length > 0;
+
+    if (hasMetricResultsError && this.flags.expectations) {
+      throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
```
pedro93: I don't agree with this approach. Just because a particular metric failed, it does not mean that the rest of the metrics should not be shown. You should return the output as is currently implemented (both the graphics and json output) with an appropriate error process code if any expectation was not meet.

willmendesneto: I got it. Moving it back for run after commands, as it was before

lib/index.ts
willmendesneto: Thanks for the feedback. I'm simplifying the method and adding some types for `this.expectations` because it changes between `ExpectationMetrics` and `NormalizedExpectationMetrics` in the constructor.

willmendesneto:
@paulirish @pedro93 Could you guys take a look again, please? Thanks!

willmendesneto:
Sorry, folks. Any updates?

pedro93:
The change looks good, will you address comment https://github.com/paulirish/pwmetrics/pull/128#discussion_r126884062 aswell? I still don't like the idea of executing runs after an expectation has failed.

willmendesneto:
@pedro93 About the comment, the changes were applied in the commit https://github.com/paulirish/pwmetrics/pull/128/commits/4d878339ebcade3925afbc6df8a186f6bd3f3c31 and the `MetricsResults.error` was removed. This approach covers yours and @paulirish suggestions.

If you have any other concerns could you clarify those, please?

pedro93:
From my point of view it looks good 👍, @paulirish what about you?

lib/index.ts
```diff
@@ -86,9 +94,26 @@ class PWMetrics {
...
+  resultHasExpectationErrors(metrics: MetricsResults): boolean {
+    const expectationErrors = metrics.timings.filter((timing: Timing) => {
+      const expectation = this.expectations[timing.id];
+      if (!expectation) {
+        return false;
+      }
+      const expectedErrorLimit = expectation.error;
+      return expectedErrorLimit !== undefined && timing.timing >= expectedErrorLimit;
+    });
+    return expectationErrors.length > 0;
```
rafaelverger: You could user `return metrics.timings.some(testFn)`

test/index.js
```diff
@@ -116,6 +116,20 @@ describe('PWMetrics', () => {
         });
       });
     });
+
+    describe('with one run with expectations', () => {
+      beforeEach(() => {
+        const {opts} = runOptions.startWithOneRunWithExpectations;
+        pwMetrics = new PWMetrics(runOptions.startWithOneRun.url, opts);
```
rafaelverger: You should use `url` from the same fixture here, right? `startWithOneRunWithExpectations`

lib/index.ts
```diff
@@ -86,9 +94,26 @@ class PWMetrics {
...
   }
 
+  resultHasExpectationErrors(metrics: MetricsResults): boolean {
+    const expectationErrors = metrics.timings.filter((timing: Timing) => {
+      const expectation = this.expectations[timing.id];
+      if (!expectation) {
+        return false;
+      }
+      const expectedErrorLimit = expectation.error;
+      return expectedErrorLimit !== undefined && timing.timing >= expectedErrorLimit;
```
rafaelverger: Is there any parsing step to `expectation.error`? In the fixture's tests you'd defined it as `'>=3000', but here you've hardcoded the operator and are using the `error` as a number, is it correct?

lib/index.ts
```diff
@@ -57,17 +59,23 @@ class PWMetrics {
...
     let metricsResults: MetricsResults[] = [];
 
+    let resultHasExpectationErrors = false;
+
     for (let runIndex of runs) {
       try {
-        metricsResults[runIndex] = await this.run();
+        const currentMetricResult: MetricsResults = await this.run();
+        if (!resultHasExpectationErrors) {
+          resultHasExpectationErrors = this.resultHasExpectationErrors(currentMetricResult);
```
rafaelverger: It would be better if `this.resultHasExpectationErrors` threw an error when `this.flag.expectations` is on, don't u think? This way, expectation errors would be handled like other errors and this `this.run` execution would be flagged as `FAILED_RUN` instead of `SUCCESS_RUN`

lib/index.ts
willmendesneto: Take a look in [NormalizedExpectationMetrics](https://github.com/paulirish/pwmetrics/blob/5ed7d0f92cb294c23020638bc54160609f7cedad/types/types.ts#L110) interface and [where it's used in the code](https://github.com/paulirish/pwmetrics/pull/128/files#diff-8ad51fb63d3211deee9a805a512386adR23)

test/index.js
willmendesneto: Not in that case, because if you look into the fixtures, there are some cases covering when the configuration has expectations without configuration. That was the reason to create a new one instead use an existent one.

lib/index.ts
willmendesneto: Not, because this code should cover when `runs` is more than `1` via configuration, so the code should run and, after that, should be validated, without the circuit breaker.

test/index.js
willmendesneto: Oh! I got that. Updating right now. Thanks for the feedback.

lib/index.ts
rafaelverger: I've got it... just seems weird that you can set a threshold if an operator that will be ignored. Not a problem of this PR anyway..

lib/index.ts
```diff
@@ -86,9 +94,25 @@ class PWMetrics {
         await sheets.appendResults(results.runs);
       }
     }
+
+    if (resultHasExpectationErrors && this.flags.expectations) {
+      throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
```
denar90: @paulirish didn't we wanna CI continue its work and don't throw on found expectations?

denar90:
ok, so land it as it is and improve stuff in future pr's.
@willmendesneto thanks for the contribution 👍 and using pwmetrics :)

willmendesneto:
Thanks @denar90 @pedro93 @paulirish @rafaelverger. It will be great if you guys can create issues with those improvements so that we can try to improve those too.

jooj123:
thanks @willmendesneto  🎉

lib/index.ts
paulirish: Yes I think the CLI should catch this exception and continue going. 

But IMO it makes sense that the node module throws. (Consider anyone doing a `require('pwmetrics')`)..

I'm open to alternatives tho.

jooj123:
@denar90 any chance of getting a release published (we need these changes at the moment)
thanks

denar90:
Hi, folks. Done! 💥

PR #131, opened 2017-07-18
Feature/add devtools viewer link to sheets

david-benson-fc:
If files are uploaded to drive, then link to chrome devtools timeline viewer which opens the given file is added to comment column of google sheets file.
Requires app authorisation.

This is a quick and dirty hack - am open to advice for improvements or how to write tests for it, or am happy for it to be considered a feature request.

Currently there is no way to connect a sheets entry and a file - files are identified by a unix timestamp which is not present in the sheet, and even if it were, x-referencing would be tedious.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -174,6 +174,10 @@ class PWMetrics {
 
       if (this.flags.upload) {
         const driveResponse = await upload(data, this.clientSecret);
+
+        preparedData.fileId = driveResponse.id;
+        preparedData.fileName = driveResponse.name;
+
         this.view(driveResponse.id);
       }
 
--- a/lib/sheets/index.ts
+++ b/lib/sheets/index.ts
@@ -13,6 +13,7 @@ const metricsIds = require('../metrics').ids;
 const SHEET_TYPES = {
   'GOOGLE_SHEETS': 'GOOGLE_SHEETS'
 };
+const VIEWER_URL_PREFIX = 'https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=drive://';
 
 class Sheets {
   constructor(public config: SheetsConfig, public clientSecret: AuthorizeCredentials) {
@@ -50,6 +51,12 @@ class Sheets {
     results.forEach(data => {
       const getTiming = (key: string) => data.timings.find(t => t.id === key).timing;
       const dateObj = new Date(data.generatedTime);
+      let viewerUrl = '';
+
+      if(typeof data.fileName !== 'undefined' && typeof data.fileName !== 'undefined') {
+          viewerUrl = VIEWER_URL_PREFIX + data.fileId;
+      }
+
       // order matters
       valuesToAppend.push([
         data.lighthouseVersion,
@@ -62,7 +69,8 @@ class Sheets {
         getTiming(metricsIds.VC100),
         getTiming(metricsIds.TTFI),
         getTiming(metricsIds.TTCI),
-        getTiming(metricsIds.VC85)
+        getTiming(metricsIds.VC85),
+        viewerUrl
       ]);
     });
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ matrix:
   include:
     - node_js: "6"
     - node_js: "7"
-    - node_js: "7"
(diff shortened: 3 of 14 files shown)


denar90:
Hi. Thanks for the contribution.
I have a concern about this one because it becomes unpredictable when data is submitted to sheets with or without timeline viewer link.
So in case
 `pwmetrics https://example.com --upload --submit` will add column with link
and
`pwmetrics https://example.com --submit` will send data as it is right now
Right now I don't have any idea how to make it more predictable and obvious...

cc @paulirish @pedro93 , maybe @samccone 
any ideas guys or you just ok with this behavior?

lib/sheets/index.ts
```diff
@@ -50,6 +51,12 @@ class Sheets {
     results.forEach(data => {
       const getTiming = (key: string) => data.timings.find(t => t.id === key).timing;
       const dateObj = new Date(data.generatedTime);
+      let viewerUrl = '';
+
+      if(typeof data.fileName !== 'undefined' && typeof data.fileName !== 'undefined') {
```
willmendesneto:
```javascript
if(data.fileName) {
  viewerUrl = VIEWER_URL_PREFIX + data.fileId;
}
```

david-benson-fc: Huh - just noticed one of those is supposed to be data.fileId, but that's the only one needed now - I originally intended to use both, but ditched the idea of using the file name as the link text.

I use `typeof` rather than a truthy check as it's won't choke on undefined.

david-benson-fc:
@denar90 Ah good point - I hadn't considered a case where someone uploaded to drive with out submitting to sheets! Well spotted.

david-benson-fc:
@denar90 How's this?

https://github.com/paulirish/pwmetrics/pull/131/commits/cddb41241940044b82fdfc6465c871f3ef6f6dd0

lib/sheets/index.ts
willmendesneto: So you don't need to be concerned of that because `undefined` is `false` when type checked in a boolean scenario.

DisasterMan78:
@denar90 That makes sense - it's certainly tidier. And yes, documentation is handy!

I've got some urgent work today, I'll take a look at this later.

(I've switched accounts btw - the comments before were from my work account on my previous contract)

DisasterMan78:
@denar90 Excuse the delay, life likes to interfere.

I've created a function `configIsSubmitAndUpload()` (bad name, am open to suggestions) in lib/index.ts, which can take advantage of your `getTimelineViewerUrl()` const (didn't spot that before!) but the `Sheets` class can't access it as it's out of scope. I don't want to be passing it in to `Sheets`, but that leaves me a little stuck.

I've just done it for now, but I don't like it much. Overall it's much more concise and tidy though.

Any suggestions?

Also, I haven't been able to properly check the viewer link as the viewer refuses to acknowledge it's authorised to access my drive account. Grr...

denar90:
@DisasterMan78 sorry for so big delay. I'll take a look at it on the upcoming weekend.

lib/index.ts
```diff
@@ -96,7 +96,7 @@ class PWMetrics {
         console.log(messages.getMessage('MEDIAN_RUN'));
         this.displayOutput(results.median);
       } else if (this.flags.submit) {
-        const sheets = new Sheets(this.sheets, this.clientSecret);
+        const sheets = new Sheets(this.sheets, this.clientSecret, this.configIsSubmitAndUpload);
```
denar90: What if not pass function to the sheets but extend `results.runs` with `viewerUrl` like
```
const sheets = new Sheets(this.sheets, this.clientSecret, this.configIsSubmitAndUpload);
await sheets.appendResults(extendResultsToSubmit(results.runs));

....
//extendResultsToSubmit method
return results.map(data => {
  if (this.flags.submit && this.flags.upload) {
    data.viewerUrl = getTimelineViewerUrl(data.fileId);
  }
  return data;
});
```
in this case we can cut off some stuff
wdyt?

DisasterMan78:
Quick update on this: The contract I was on ended when I was working on this so never got to finish it off (hence the different GH account). It's still on my list to return to when I find the time, but if anyone would like to have a crack at it in the meantime, feel free.

PR #133, opened 2017-07-23
Adding missed types

willmendesneto:
Adding missed types in some methods/functions. These changes are covering:

- `lib/expectations.ts`;
- `lib/index.ts`;
- `lib/sheets/gsheets.ts`;
- `lib/utils/messages.ts`;


samccone:
Hi, the typescript compiler is quite smart! These can all be infered and thus are not need.

PR #134, opened 2017-07-24
Adds chromePath option to enable specifying specific Chrome to run.

natedanner:


lib/index.ts
```diff
@@ -182,7 +182,8 @@ class PWMetrics {
       console.log(messages.getMessage('LAUNCHING_CHROME'));
       this.launcher = await launch({
         port: this.flags.port,
-        chromeFlags: this.flags.chromeFlags
+        chromeFlags: this.flags.chromeFlags,
+        chromePath: this.flags.chromePath
```
denar90: @samccone does chrome-launcher normalizing path under the hood or we should do it?

natedanner: chrome-launcher doesn't do any normalizing, it passes the provided string as the path to spawn Chrome as a child process. https://github.com/GoogleChrome/lighthouse/blob/master/chrome-launcher/chrome-launcher.ts#L174 I can add that.

Issue #132, opened 2017-07-20
Audit error: trace not at least 5 seconds longer than FMP Audit key: first-interactive


denar90:
### Config / CLI options
`pwmetrics https://example.com`

### Steps to reproduce.
run `pwmetrics https://example.com`

### Stack trace
```bash
Launching Chrome
  ✘ Error: Unable to complete run 1 of 1 due to Audit error: trace not at least 5 seconds longer than FMP Audit key: first-interactive.
```

### Environment
1. `pwmetrics` version: 3.0.1
2. Chrome version: Version 61.0.3162.0 (Official Build) canary (64-bit)
3. OS version: MacOs Sieera 10.12.4

staabm:
same on 
  pwmetrics 3.1.0
  win10
  Chrome 62.0.3169.0
  node 8.2.1

denar90:
Looks like LH is not providing metrics we needed.
Our logic is dumb, get metrics -> draw chart, if there are no metrics -> throw error.
I'm wanna change it to - get metrics -> draw chart, if there are no metrics -> draw chart but with missed metrics.
I have one problem - time. Really really hope to look at it on a weekend.
Today is #opesourcefriday but not at my company...

pedro93:
@denar90 and @staabm can you reproduce this consistently? I've been trying but pwmetrics works fine for me.

Pwmetrics version: 3.1.0 (from master)
Chrome version: 60.0.3112.78
OS: Ubuntu 16.04 LTS
NodeJS: v8.2.1

denar90:
Yup. Still have this issue consistently...

Issue #135, opened 2017-07-24
No domContentLoaded event in trace Audit key: first-interactive.


epodgaetskiy:
pwmetrics https://gdesport.info

 Error: Unable to complete run 1 of 1 due to Audit error: No domContentLoaded event in trace Audit key: first-interactive.

Can you help me about this error?

denar90:
It's related to https://github.com/paulirish/pwmetrics/issues/132 
Looks like some metrics results are misssed. 
I'll try to debug Lighthouse and get back with results.

pedro93:
@epodgaetskiy What are you system's characteristics? Node, pwmetrics, chrome and OS version?

I managed to run it fine:
![screenshot from 2017-07-29 10-48-36](https://user-images.githubusercontent.com/3718654/28743828-82b01c26-744b-11e7-8575-0902f9418487.png)

denar90:
`node v8.2.1`
`npm 5.3.0`
`chrome - latest canary`

PR #136, opened 2017-07-28
check expectation errors then feature flag enabled

denar90:
fixes https://github.com/paulirish/pwmetrics/issues/132 but not https://github.com/paulirish/pwmetrics/issues/135


denar90:
nope....

PR #145, opened 2017-11-22
Bump lighthouse version to 2.7.0

denar90:


denar90:
Ok, so since LH 2.7.0 dropped ts I moved types to repo.

PR #147, opened 2018-01-06
fix/add chrome flags transformation

denar90:
turned out that chrome flags are not passed properly to chrome launcher. this pr fix the problem

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -68,6 +68,8 @@ class PWMetrics {
         this.expectations = expectations.normalizeMetrics(this.expectations);
       } else throw new Error(messages.getMessageWithPrefix('ERROR', 'NO_EXPECTATIONS_FOUND'));
     }
+
+    this.flags.chromeFlags = parseChromeFlags(this.flags.chromeFlags);
   }
 
   async start() {
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -3,6 +3,7 @@
 
 import {launch, LaunchedChrome} from 'chrome-launcher';
 const lighthouse = require('lighthouse');
+const parseChromeFlags = require('lighthouse/lighthouse-cli/run').parseChromeFlags;
 const perfConfig: any = require('./lh-config');
 const opn = require('opn');
 const path = require('path');
@@ -39,7 +40,7 @@ class PWMetrics {
     view: false,
     expectations: false,
     json: false,
-    chromeFlags: []
+    chromeFlags: ''
   };
   runs: number;
   sheets: SheetsConfig;
@@ -188,7 +189,7 @@ class PWMetrics {
       console.log(messages.getMessage('LAUNCHING_CHROME'));
       this.launcher = await launch({
         port: this.flags.port,
-        chromeFlags: this.flags.chromeFlags,
+        chromeFlags: parseChromeFlags(this.flags.chromeFlags),
         chromePath: this.flags.chromePath
       });
       this.flags.port = this.launcher.port;
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -38,7 +38,7 @@ class PWMetrics {
     upload: false,
     view: false,
     expectations: false,
-    output: false,
+    json: false,
     chromeFlags: []
   };
   runs: number;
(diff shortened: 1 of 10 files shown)


lib/index.ts
```diff
@@ -39,7 +40,7 @@ class PWMetrics {
     view: false,
     expectations: false,
     json: false,
-    chromeFlags: []
+    chromeFlags: ''
```
denar90: another thing, I want to add `--headless --no-sandbox --disable-setuid-sandbox` by default in another pr. wdyt?

lib/index.ts
```diff
@@ -3,6 +3,7 @@
 
 import {launch, LaunchedChrome} from 'chrome-launcher';
 const lighthouse = require('lighthouse');
+const parseChromeFlags = require('lighthouse/lighthouse-cli/run').parseChromeFlags;
```
pedro93: This parseChromeFlags method documentation says it is available an as export for testing purposes, see:
https://github.com/GoogleChrome/lighthouse/blob/a2bcf82c13e18a5591e582e6f5760455783b9bb3/lighthouse-cli/run.js#L27

Is Lighthouse using another method to pass on flags to chrome or this it just a documentation typo?
If the former then this change may work for now, but it won't be future-proof

lib/index.ts
pedro93: What is the objective? 
Faster overall pwmetrics execution? I'm concerned the reported values will not match what a developer can manually get.
Last time I tried headless lighthouse performance tests the results were quite different from normal execution and not what I found via chrome dev tools with an empty cache.

lib/index.ts
denar90: Yeah, just wanted to avoid copy/pasta since LH and pwmetrics tightly coupled

>Is Lighthouse using another method to pass on flags to chrome or this it just a documentation typo?

I'm not sure.

For preventing problems in future I see this scenario:
1, of course, add test for this case and use it as it is
2. fork LH and move this method to utilities for safe using

lib/index.ts
denar90: I'm running all env in ubuntu docker container and could not run Chrome without `--headless --no-sandbox --disable-setuid-sandbox`...

@paulirish have either you or team ever experienced different results comparing between headless run and regular one?

paulirish: > @paulirish have either you or team ever experienced different results comparing between headless run and regular one?

--headless used to not support network throttling at all, so the perf numbers were indeed VERY different.
now it DOES. yay!

so the numbers will be closer. it won't be doing some painting/gpu work that headful usually does, but should be roughly comparable.

that said, we haven't switched to using headless for any performance sensitive measurement yet.  just haven't spent the time to confirm the numbers are good enough. :)

lib/index.ts
pedro93: I see 3 options with regards to this:
1.) Either use the chrome flags as LH does, as a simple string and pass that over to chrome launcher directly.
2.) Pull Request LH to make that particular function part of the API (would work but I think it's ugly and breaks separation of concerns).
3.) Pull request chrome launcher and address the issue there. Perhaps enabling chrome launcher to take an array of strings.

lib/index.ts
denar90: Ok, so let's just document this case for now, without adding those flags by default.

lib/index.ts
denar90: >Either use the chrome flags as LH does, as a simple string and pass that over to chrome launcher directly.

@pedro93 bug was exactly about that. right now doing like `pwmetrics https://example.com --chrome-flags="--headless --no-sandbox --disable-setuid-sandbox"` didn't pass expected params to chrome launcher. Without parsing `chromeFlags` will be passed as string `--headless --no-sandbox --disable-setuid-sandbox` to ChromeLauncher.
Since we are using LH and run ChromeLauncher manually we have to do the same chrome flags adaptation as LH does. 

Sorry that I didn't explain that in a PR description

> Pull request chrome launcher and address the issue there. Perhaps enabling chrome launcher to take an array of strings.

It does that https://github.com/GoogleChrome/chrome-launcher/blob/v0.10.2/src/chrome-launcher.ts already

pedro93: Sorry I didn't reply sooner, I now understand the issue. I think this is ok for now but we must be aware that it will be a hidden dependency on a utility LH method.

paulirish: Yah seems fine. Even if it changes in the future, it will be exposed. So, 👍

denar90:
I'd better add some tests on this weekend and then we can merge this one too.
And make release :)

PR #169, opened 2018-04-04
Patch/update docs

jooj123:
The format of the sheets clientSecret wasnt really clear from the docs - so just updating it here


denar90:
thanks 👍

PR #178, opened 2018-05-28
updates for using lighthouse v3

denar90:
Core of changes to use LH v3
- updated perf config
- cleaned up metrics adapter
- cleaned types

Has to be merged after final release and test of v3 version.


lib/index.ts
```diff
@@ -1,15 +1,15 @@
...
 // Licensed under the Apache License, Version 2.0. See LICENSE
 
-import {launch, LaunchedChrome} from 'chrome-launcher';
-const lighthouse = require('lighthouse');
+import { launch, LaunchedChrome } from 'chrome-launcher';
+const lighthouse = require('lighthouse/lighthouse-core');
 const parseChromeFlags = require('lighthouse/lighthouse-cli/run').parseChromeFlags;
-const perfConfig: any = require('./lh-config');
+const perfConfig: any = require('lighthouse/lighthouse-core/config/plots-config.js');
+perfConfig.extends = 'lighthouse:default';
```
denar90: note: already fixed in LH master branch and has to be removed

denar90:
it will be nice if someone takes a look at it before further work on it...
thanks)

lib/index.ts
```diff
@@ -1,15 +1,15 @@
 // Copyright 2016 Google Inc. All Rights Reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE
 
-import {launch, LaunchedChrome} from 'chrome-launcher';
-const lighthouse = require('lighthouse');
+import { launch, LaunchedChrome } from 'chrome-launcher';
+const lighthouse = require('lighthouse/lighthouse-core');
 const parseChromeFlags = require('lighthouse/lighthouse-cli/run').parseChromeFlags;
-const perfConfig: any = require('./lh-config');
+const perfConfig: any = require('lighthouse/lighthouse-core/config/plots-config.js');
```
paulirish: I think you should just provide a config file here in this repo. We don't have much reason to maintain that plots-config fwiw. :)

### `metrics` audit
while everything you need is in the new audit called `metrics` it doesnt come with human-readable names.  Since you provide those in `messages.ts` maybe its fine and all you really need is `metrics`. I think this could work.


image


### individual audits
If you really need all the individual audits:

```js

module.exports = {
  extends: 'lighthouse:default',
  settings: {
    onlyAudits: defaultConfig.categories.performance.auditRefs
	.filter(a => a.group && a.group === 'metrics')
	.map(a => a.id)
  },
};


```

lib/metrics.ts
```diff
@@ -51,49 +32,34 @@ function prepareData(res: LighthouseResults): MetricsResults {
...
 
   return {
     timings,
-    timestamps,
-    generatedTime: res.generatedTime,
+    generatedTime: res.fetchTime,
     lighthouseVersion: res.lighthouseVersion,
-    initialUrl: res.initialUrl,
-    url: res.url
+    initialUrl: res.finalUrl,
```
paulirish: initial is final? 

i'd recommend adopting `requestedUrl` and `finalUrl` names for this metrics export.

lib/utils/messages.ts
```diff
@@ -44,24 +44,16 @@ const getMessage = function (messageType: string, ...args: any[]) {
...
       return 'Perceptual Speed Index';
-    case 'fv':
-      return 'First Visual Change';
-    case 'vc':
-      return 'Visually Complete 100%';
-    case 'tti':
+    case 'estimated-input-latency':
+      return 'Estimated Input Latency';
+    case 'first-cpu-idle':
+      return 'Time to First Interactive';
```
paulirish: the metric is renamed so it'd be best to call it "First CPU Idle" now

lib/utils/messages.ts
```diff
@@ -44,24 +44,16 @@ const getMessage = function (messageType: string, ...args: any[]) {
       return `No matching message prefix: ${args[0]}`;
     case 'METRIC_IS_UNAVAILABLE':
       return `Sorry, ${args[0]} metric is unavailable`;
-    case 'ttfcp':
+    case 'first-meaningful-paint':
```
paulirish: seems wrong.

in 3.0 FCP is both in 'metrics' audit and its own. i would def recommend showing **both** FCP and FMP in pwmetrics. (plots doesnt have FCP)

lib/index.ts
denar90: Paul, thanks. Would you like to keep all metrics we already have or reduce their number? I'm kinda confused because I faced that people just ignore FCP and rely only on FMP in perf aduits/case studies etc. As result with 3.0 and plots-config I thought maybe trends were changed...

denar90:
at list tests passed ;)

lib/metrics/metrics.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE
+
+export default {
```
denar90: I see this one useful for people using expectations and stuff. What they need is just import our constants and use them for expectations config.
```ts
const METRICS = require('pwmetrics/lib/metrics-adapter/metrics');

module.exports = {
  url: 'http://my-awesome-site.com'
  flags: {
    expectations: true
  },
  expectations: {
    [METRICS. TTFCP]: {
       warn: '>=1000',
       error: '>=2000'
    }
  }
}
```

 Other then that cleaner for our code to have map metrics ids.

lib/index.ts
paulirish: The way we look at the problem now is:

### key metrics

* FCP
* Speed Index
* TTI

### worthwhile diagnostic metrics

* FMP
* FCPUIdle

EIL is interesting but could use some work. it'll also change substantially in the future. And it's not on the same timescale as the others.

FirstVC should mostly be duplicated by FCP.

VizComplete is cool unless it marked the wrong thing, in which case it and Speed Index are totally useless. :)

denar90: @paulirish  but VizComplete is not part of metrics.
```js
auditRefs: [
        {id: 'first-contentful-paint', weight: 3, group: 'metrics'},
        {id: 'first-meaningful-paint', weight: 1, group: 'metrics'},
        {id: 'speed-index', weight: 4, group: 'metrics'},
        {id: 'interactive', weight: 5, group: 'metrics'},
        {id: 'first-cpu-idle', weight: 2, group: 'metrics'},
        {id: 'estimated-input-latency', weight: 0, group: 'metrics'},
```
Should it be added there?
Or,  should we use `pwmetrics-events` and get it manually?

denar90:
@paulirish we are good to go after solving https://github.com/paulirish/pwmetrics/pull/178#discussion_r199360132
Then I'll do manual regression and update docs.

Since LH 3.0 is done, It will be nice to announce pwmetrics updates right after LH announcement.

denar90:
@paulirish can you take a look here  👀?

denar90:
Opened questions are:
- handle vizcomplete 
- gulp example still failes on CI :( (we can live with that for now)
- code review

package.json
```diff
@@ -17,12 +17,12 @@
...
   "dependencies": {
-    "@gribnoysup/wunderbar": "^2.2.0",
-    "@types/node": "^6.0.45",
+    "@gribnoysup/wunderbar": "^2.2.1",
+    "@types/node": "^10.1.3",
     "chrome-launcher": "^0.10.2",
     "google-auth-library": "^0.10.0",
     "googleapis": "^16.0.0",
-    "lighthouse": "^2.9.4",
+    "lighthouse": "3.0.0",
```
paulirish: you can bump it to 3.0.3 now

package.json
```diff
@@ -32,11 +32,12 @@
     "eslint": "^3.12.0",
     "eslint-config-google": "^0.7.1",
     "gulp": "^3.9.1",
-    "gulp-connect": "^5.0.0",
+    "gulp-connect": "^5.5.0",
     "mocha": "^3.2.0",
     "sinon": "^1.17.7",
     "sinon-chai": "^2.8.0",
-    "typescript": "^2.0.3"
+    "typescript": "2.9.1-insiders.20180516",
```
paulirish: this can go to 3.0.0 now too :)

package.json
```diff
@@ -32,11 +32,12 @@
...
+    "vscode-chrome-debug-core": "3.23.8"
```
paulirish: not sure why this is added.  was it types?

.travis.yml
```diff
@@ -1,25 +1,28 @@
...
 dist: trusty
 cache:
   yarn: true
   directories:
     - node_modules
-before_install:
-  - npm install -g --no-progress yarn@v0.22.0
+    - lighthouse-extension/node_modules
+    - lighthouse-viewer/node_modules
+    - /home/travis/.rvm/gems/
```
paulirish: the two lighthouse-* folders dont make sense to me here.
and i dont think you're running ruby, so you can nuke this one too.

lib/index.ts
```diff
@@ -107,15 +98,15 @@ class PWMetrics {
     }
 
     if (resultHasExpectationErrors && this.flags.expectations) {
-      throw new Error(messages.getMessage('HAS_EXPECTATION_ERRORS'));
+      console.log(getMessage('HAS_EXPECTATION_ERRORS'));
```
paulirish: you want to let things continue in this case? im fine with that, just pointing out that the behavior is changing. :)
probably should console.error() if its staying non-throwing.

lib/metrics/metrics-adapter.ts
```diff
@@ -0,0 +1,67 @@
...
+  checkAudits(audits);
+
+  const colorP0 = 'yellow';
+  const colorP2 = 'green';
+  const colorVisual = 'blue';
+
+  const timings: Timing[] = [];
+
+  Object.keys(audits).forEach(metricKey => {
+    const metric = audits[metricKey];
```
paulirish: the `audits['metrics']` audit contains everything i think you want. up to you, but it would be slightly cleaner to pull from just that.

https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/metrics.js#L96-L138

btw VizCommplete is in there as `observedLastVisualChange`. however that's **not** adjusted if run with simulated throttling aka lantern.

lib/metrics/metrics.ts
```diff
@@ -0,0 +1,11 @@
...
+  TTFCP: 'first-contentful-paint',
+  TTFMP: 'first-meaningful-paint',
+  TTF_CPU_IDLE: 'first-cpu-idle',
+  TTCI: 'interactive',
+  PSI: 'speed-index',
+  VISUALLY_COMPLETE: 'vc100',
```
paulirish: seems like this isn't being reported. 

While it's *possible* to get the number, you'll have to choose if you want to go with devtools throttling just to get this.

recipes/gulp/gulpfile.js
```diff
@@ -57,22 +39,19 @@ const runPwmetrics = function() {
  * @param {Object} results - Pwmetrics results obtained through Lighthouse
  */
 const handleOk = function(results) {
-  stopServer();
-  console.log(results);
+  console.log(JSON.stringify(results, null, 4));
```
paulirish: 4 -> 2.   😝

types/types.ts
```diff
@@ -53,127 +29,86 @@ interface FeatureFlags {
...
-  5: number; // PSI timing
-  6: number; // FV timing
-  7: number; // VC100 timing
-  8: number; // TTI timing
-  9: number; // VC85 timing
+  3: number; // TTFMP timing
+  4: number; // PSI timing
+  5: number; // VISUALLY_COMPLETE timing
+  6: number; // TTFI timing
+  7: number; // TTCI timing
```
paulirish: since these are changing position, i guess these sheets will be a bit broken now.

lib/metrics/metrics.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE
+
+export default {
+  TTFCP: 'first-contentful-paint',
+  TTFMP: 'first-meaningful-paint',
+  TTF_CPU_IDLE: 'first-cpu-idle',
+  TTCI: 'interactive',
```
paulirish: TTI

lib/metrics/metrics.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2018 Google Inc. All Rights Reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE
+
+export default {
+  TTFCP: 'first-contentful-paint',
+  TTFMP: 'first-meaningful-paint',
+  TTF_CPU_IDLE: 'first-cpu-idle',
```
paulirish: TTFCPUIDLE  ?   :)

lib/metrics/metrics.ts
```diff
@@ -0,0 +1,11 @@
...
+  TTCI: 'interactive',
+  PSI: 'speed-index',
```
paulirish: PSI -> SI everywhere. we ditched perceptual speed index and now just calculate speed index.

lib/metrics/metrics.ts
denar90: yup, just dummy placed here becase of confusion with vizComplete. 
wdyt if we call it just like it's done in LH to not involve extra effort into `you'll have to choose if you want to go with devtools throttling just to get this.` ?

package.json
denar90: I think it's fixed in LH now, so can be dropped 👍

types/types.ts
denar90: nice catch 👍

package.json
denar90: hmm, still needed because without that tsc build failed

lib/metrics/metrics-adapter.ts
denar90: So to clarify, if we debug LH method `_runAudit` in runner.js, while we call 
`const product = await audit.audit(artifacts, {options: auditOptions, settings: settings});`

`artifacts` has all stuff needed for `Metrics.audit` method, 
but if we will look at artifacts which LH provide after run we will see mostly the same object except all needed methods like `requestCriticalRequestChains`, `requestEstimatedInputLatency`, `requestFirstCPUIdle` and many others. If I'm not wrong it's ComputedArtifacts.

That's why call `Metrics.audit` will crash on lines https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/metrics.js#L34-L41

Maybe it's a bug and LH has also provide artifacts with all necessary methods, so they can be used in `Metrics.audit` ?

package.json
paulirish: huh. weird. we're about to remove that dependency: https://github.com/GoogleChrome/lighthouse/pull/5836

lib/metrics/metrics-adapter.ts
paulirish: Hmmmmmmmm

This is getting into some Lighthouse internals stuff that I don't think pwmetrics needs to worry about, but maybe I'm missing something. I'm saying the config will be have `onlyAudits: ['metrics']`. No more filtering to the metric group ids.. Just pulling this singular audit. By including it, it should compute all the metrics.

`adaptMetricsData` still gets passed `data.lhr`. But you'll pull numbers out like:

```js
lhr.audits.metrics.details[metricName]
lhr.audits.metrics.details[`${metricName}Ts`]
```

lib/metrics/metrics.ts
paulirish: sure, but right now i think there's no configuration option for pwmetrics to use devtools throttling instead of the default simulation. that'll be changes to your config. 

might want to delay this and do it later, since it adds some complexity.

lib/metrics/metrics-adapter.ts
denar90: sorry for the delay,
yup it looks more reasonable:) I just got confused and dug in wrong diraction 💫 
hope to get back to it later today)

package.json
denar90: probably we will wait for PR to be merged...

lib/metrics/metrics.ts
denar90: ok, dropping it then.
will you be ok with just 5 metrics?

denar90:
Finally CI passes 🎉 
So waiting for LH  to be updated with removing vscode-chrome-debug-core.
Other than that if we are ok to go without vizComplete metric I'll do final manual check and ready to 🚢  P.S. I still don't know what version it will be. Looks like v4...

package.json
denar90: dropped 💥 
just waiting for next LH release

denar90:
@paulirish we are ready, I think)

denar90:
I'm going to merge it at the and of the day today. In case of issues/comments, we can fix them in separate PR.

paulirish:
Sounds good

On Sep 12, 2018 1:53 AM, "Artem Denysov" <notifications@github.com> wrote:

I'm going to merge it at the and of the day today. In case of
issues/comments, we can fix them in separate PR.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/paulirish/pwmetrics/pull/178#issuecomment-420566883>,
or mute the thread

denar90:
Merged 🎉

denar90:
Still wondering about version,
3.4.0 or 4.0.0 wdyt?

denar90:
>but since pwmetrics is using simulated throttling instead of devtools (right?), then the metrics numbers aren't really comparable. that'd be a breaking change. :)

yup. thanks)

Issue #184, opened 2018-08-01
Hide console output


santoshjoseph99:
When using the API, I think it would be good idea not to output anything to the console, since you get that data from the `start` call.
Seems pretty simple, just pass a flag into the `options` object to silence the `console.log`'s.
I can get this done if needed.

output when using API:
<img width="751" alt="screen shot 2018-07-31 at 9 29 40 pm" src="https://user-images.githubusercontent.com/461049/43501135-ea3f2542-9508-11e8-9eb1-dbac8082f377.png">

### Environment
1. `pwmetrics` version:
2. Chrome version:
3. OS version:

denar90:
I think we can use flag her, aka `showLaunchingResults` which might be true by default. wdyt?

santoshjoseph99:
sounds good. I'll submit a PR.

denar90:
fixed via #185

PR #185, opened 2018-08-21
feat: showOutput option & log function

santoshjoseph99:
Related to https://github.com/paulirish/pwmetrics/issues/184

Test file:
https://gist.github.com/santoshjoseph99/c796663002b03812cc197fe1f8bb8d38

Note:
1. I could add a parameter to `logFunc` to check if the message is a `log` or an `error`, since right now I'm just using `console.log`.
2. I could also set this up so that `logFunc` can be passed into `PWMetrics` as part of the `options`, so it gives more control to the caller on how to show the messages.

### DIFF
--- a/lib/chart/chart.ts
+++ b/lib/chart/chart.ts
@@ -3,7 +3,7 @@ const eol = require('os').EOL;
 
 import {Timing, ChartOptions} from '../../types/types';
 
-const drawChart = (timings: Timing[], options: ChartOptions) => {
+const drawChart = (timings: Timing[], options: ChartOptions, logFunc: any) => {
   const {lmargin, width, xlabel, xmin, xmax} = options;
 
   const normalizedTimings = timings.map(value => {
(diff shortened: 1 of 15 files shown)


santoshjoseph99:
@denar90 sounds good.  I'll update.

santoshjoseph99:
@denar90  I'm not sure what you mean by the land on the docs?

Also I'm seeing a failure in CI...should I change the signature of the log function to be:
```
  log(msg: any, ...args: any[]) {
    if(Logger.options.showOutput){
      console.log(msg, ...args);
    }
  }
```
?

denar90:
Moving thread here,
---
>lgtm 👍
one nitpick, can you update docs so we can land it?
---
> @denar90 I'm not sure what you mean by the land on the docs?
Also I'm seeing a failure in CI...should I change the signature of the log function to be:
```
  log(msg: any, ...args: any[]) {
    if(Logger.options.showOutput){
      console.log(msg, ...args);
    }
  }
```
>?
---
Sorry, for the confusion. 
I meant that it will be nice if you can update `readme.md` about new flag. As result we can publish new version.
And I'm ok with signature of the log function.

denar90:
v3.3.0 published to npm 🎉

santoshjoseph99:
🆒

PR #194, opened 2018-10-10
fix: exit code is 1 when expectations are not fulfilled

GuillaumeAmat:
Fix #193 

A bit of refact of the `start` method, in order to return 1 as the exit code when expectations have errors.

PR #195, opened 2018-10-12
refactor: outputData as a proper PWMetrics method

GuillaumeAmat:
Related to: https://github.com/paulirish/pwmetrics/pull/194#discussion_r224896044

### DIFF
--- a/lib/utils/fs.ts
+++ b/lib/utils/fs.ts
@@ -3,6 +3,8 @@
 
 import * as path from 'path';
 import * as fs from 'fs';
+import * as promisify from 'micro-promisify';
+
 import {getMessageWithPrefix} from './messages';
 import {Logger} from './logger';
 
@@ -27,11 +29,11 @@ export function getConfigFromFile(fileName: string = 'package.json') {
 }
 
 export function writeToDisk(fileName: string, data: string) {
-  return new Promise((resolve, reject) => {
+  return new Promise(async (resolve, reject) => {
     const filePath = path.join(process.cwd(), fileName);
 
     try {
-      fs.writeFileSync(filePath, data);
+      await promisify(fs.writeFile)(filePath, data);
     } catch (err) {
       reject(err);
     }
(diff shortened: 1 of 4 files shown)


lib/utils/fs.ts
```diff
@@ -17,7 +23,20 @@ function getConfigFromFile(fileName: string = 'package.json') {
...
+
 }
 
-module.exports = { getConfigFromFile };
+export function writeToDisk(fileName: string, data: string) {
+  return new Promise((resolve, reject) => {
+    const filePath = path.join(process.cwd(), fileName);
+
+    try {
+      fs.writeFileSync(filePath, data);
```
denar90: wdyt if we wrap `fs.writeFile` with `micro-promisify` (we already have it as dep) and stick to async/await here?

lib/utils/fs.ts
```diff
@@ -17,7 +23,20 @@ function getConfigFromFile(fileName: string = 'package.json') {
       return config.pwmetrics || {};
     else return config;
   } else throw new Error(`Invalid config from ${fileName}`);
-  
+
 }
 
-module.exports = { getConfigFromFile };
+export function writeToDisk(fileName: string, data: string) {
```
denar90: 😍 to see it here :)

GuillaumeAmat: 😊

lib/utils/fs.ts
GuillaumeAmat: Sure, I just pushed it :)

lib/utils/fs.ts
```diff
@@ -17,7 +25,20 @@ function getConfigFromFile(fileName: string = 'package.json') {
...
+  return new Promise(async (resolve, reject) => {
```
denar90: I meant do we need Promise here at all?
We can write inside function just
```
try { 
  const filePath = path.join(process.cwd(), fileName);
  await promisify(fs.writeFile)(filePath, data);
  logger.log(getMessageWithPrefix('SUCCESS', 'SAVED_TO_JSON', filePath));
} catch (err) {
      throw (err);
    }
```

GuillaumeAmat: Oh, ok it makes sense you're right. The promise is removed there but the whole thing has to be awaited back to the call of `outputData` line 108 of `lib/index.ts` though.

Issue #201, opened 2019-01-21
Add ability to upload median values only to sheets?


timstallmann:
Hi! 

This would seem to be a natural follow-up to #181. I'd love to see a config file value like `sheets.options.reportMedianOnly` which would only upload the median value for each metric across that batch of runs. 

I'd be down to poke in the code and try to work on a PR, if folks feel like that's a worthwhile feature?

denar90:
Hi. Looks like it's also related to #199. I'll be more than happy to have this feature.

PR's are always welcome 😄

denar90:
thx, published with v4.2.0 release

Issue #199, opened 2019-01-10
Improve data collection


denar90:
![image](https://user-images.githubusercontent.com/6231516/50973359-5f4bfc80-14f1-11e9-917c-8b3936c8c14d.png)

Sometimes we can get abnormal values for one set of e.g. 10 runs.

I think we can help users to handle this case automatically. So if for set of values we find some abnoraml values we will just replace it with median value instead. So graphics will look a bit better. 
But it should be in case most of values has normal distribution.

PR #203, opened 2019-01-22
Add option to upload median run to Google Sheets (#201 and #199?)

timstallmann:
This addresses #201 and I think #199 by adding a flag which allows configuration to choose only the median run for upload to sheets.

I'm having some issues building locally which seem to be related to `@types/node`, but did manage to work around those and test that this works.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -102,7 +102,14 @@ class PWMetrics {
       this.displayOutput(results.median);
     } else if (this.flags.submit) {
       const sheets = new Sheets(this.sheets, this.clientSecret);
-      await sheets.appendResults(results.runs);
+
+      if (this.sheets.options.uploadMedian) {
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -102,7 +102,14 @@ class PWMetrics {
       this.displayOutput(results.median);
     } else if (this.flags.submit) {
       const sheets = new Sheets(this.sheets, this.clientSecret);
-      await sheets.appendResults(results.runs);
+
+      if (this.sheets.options.uploadMedian) {
+        results.median = this.findMedianRun(results.runs);
+        await sheets.appendResults([results.median,]);
```
denar90: can we?
```suggestion
        await sheets.appendResults([results.median]);
```

timstallmann: Sure! Sorry that trailing comma is probably a python style thing I have engrained.

PR #214, opened 2019-08-03
Upgrade lighthouse version

andresz1:
This PR upgrades lighthouse to version 5 and replaces `opn` with `open` since it was [renamed](https://www.npmjs.com/package/opn). The `typing` folder of lighthouse was renamed to `types` so it is also fixed. Tested it several times and everything seems fine.

Please let me know if I should do something else and any feedback would be appreciated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: node_js
 matrix:
   include:
-    - node_js: "8"
-    - node_js: "9"
+    - node_js: "10"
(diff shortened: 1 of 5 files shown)


andresz1:
Hi @denar90! thanks for quick response. You are right! just added it to `package.json`. Let me know if something else is needed ;)

denar90:
Thank you :)

PR #230, opened 2020-05-26
Update Lighthouse to version 6

Ge11ert:
So, this is actually just the basic update to a new Lighthouse version in order to get access to the [new metrics](https://web.dev/lighthouse-whats-new-6.0/#new-metrics).

I threw out Cumulative Layout Shift metric because it has only 5% impact to final score. And also First Meaningful Paint and First CPU Idle metrics have been deprecated in LH v6 update.

There are more new stuff in new LH version which should be (re)implemented, such as timing budgets and unused js. But I think it could be done in later PRs (not mine, just speaking in common)

Any thoughts/suggestions?

### DIFF
--- a/lib/expectations.ts
+++ b/lib/expectations.ts
@@ -1,6 +1,7 @@
 // Copyright 2016 Google Inc. All Rights Reserved.
 // Licensed under the Apache License, Version 2.0. See LICENSE
 
+import {DEPRECATED_METRICS} from './metrics/metrics';
 import {Logger} from './utils/logger';
 import {getAssertionMessage, getMessageWithPrefix} from './utils/messages';
 import {Timing, ExpectationMetrics, NormalizedExpectationMetrics} from '../types/types';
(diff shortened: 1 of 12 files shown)


Ge11ert:
Hmmm. Strange fact is that local run of `npm run gulp-example` ends successfully. And Travis has some kind of error from nowhere for that.

denar90:
Wow, that's awesome update. 
I'll try to go over it tomorrow. Thanks 👍

Ge11ert:
@denar90 Hi! Any news about this PR?) It seems, no one is intending to interfere

denar90:
I'm happy to merge and release this one.

Also, we are going to deprecate `pwmetrics` in favor [Lighthouse CI](https://github.com/GoogleChrome/lighthouse-ci). 

@Ge11ert any missed features in LHCI present in pwmetrics? whats your case?

Ge11ert:
@denar90 
> @Ge11ert any missed features in LHCI present in pwmetrics? whats your case?

Actually, I chose pwmetrics some time ago for the ability of multiple runnings of lighthouse cli. This is what we'd wanted to get in our team's product project in order not to rely on single run. Because as we know there are some disturbances in lighthouse results from run to run due to network state and etc.

Speaking of LHCI, I think this is great to have some tool like that, but also I suppose, pwmetrics and LHCI serve different purposes.
Using pwmetrics as easy as to install and run npm package. And if I get it right, LighHouse CI needs to be hosted somewhere with server configuration and other stuff. Which is way more complicated, to be honest.

This is not about refusing to read the docs and give it a try, no. I just barely have rights and possibilities to host LHCI on our company's staging servers for we have many teams and quite sophisticated infrastucture.

But thanks for reminding me about Lighthouse CI, definitely need to have a closer look, in case there are some options we can use easily.

